PR #21, opened 2017-07-13
SDK: Get invitation templates from public api

marcoNipo:


Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void InitializeNfieldConnection(INfieldConnectionClient connection)
+        {
+            ConnectionClient = connection;
+        }
+
+        private string SurveyInvitationTemplatesUrl(string surveyId)
+        {
+            var result = new StringBuilder(ConnectionClient.NfieldServerUri.AbsoluteUri);
+            result.AppendFormat(CultureInfo.InvariantCulture, @"Surveys/{0}/InvitationTemplate", surveyId);
+            return result.ToString();
```
neilboyd: Should be InvitationTemplate**s**

Tests/Services/NfieldSurveyInvitationTemplatesServiceTests.cs
```diff
@@ -0,0 +1,94 @@
+﻿using Nfield.Services.Implementation;
```
neilboyd: missing file header

PR #23, opened 2017-07-27
[Story] Get invitation status from public api

marcoNipo:


Tests/Services/NfieldSurveyInviteRespondentsServiceTests.cs
```diff
@@ -120,6 +120,60 @@ private static bool FilterEquals(SampleFilter filter, string name, string op, st
...
+                UnwrapAggregateException(target.GetInvitationStatusAsync(null, null)));
+        }
+
+        [Fact]
+        public void TestGetStatusAsync_SurveyIdIsEmpty_Throws()
+        {
+            var target = new NfieldSurveyInviteRespondentsService();
+
+            Assert.Throws<ArgumentException>(() =>
+                UnwrapAggregateException(target.GetInvitationStatusAsync("  ", null)));
```
pilarodriguez: Actually, this is not testing if the survey id is empty, but if it is a white space. Replace the GetInvitationStatusAsync("  ", null) by GetInvitationStatusAsync(string.Empty, null) or rename the test

marcoNipo: true :)

PR #24, opened 2017-08-01
[Story] Monitor survey and batch invitation statuses

neilboyd:


hcsauniere:
TotalCount is an int and not a long in both classes (as in Nfield-Manager-Invitation)

PR #25, opened 2017-08-03
crud invitation templates

krakie:


Library/Services/INfieldSurveyInvitationTemplatesService.cs
```diff
@@ -25,8 +25,23 @@ namespace Nfield.Services
...
+
+        /// <summary>
+        /// Update an invitation template for the specific survey
+        /// </summary>
+        Task UpdateAsync(string surveyId, InvitationTemplateModel invitationTemplate);
+
+        /// <summary>
+        /// Delete an invitation template for the specific survey
+        /// </summary>
+        Task<bool> DeleteAsync(string surveyId, string templateId);
```
marcoNipo: The request class for deleting an invitation has templateId as int: I changed the type of the parameter for the controller method as well. Any thoughts?

Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
```diff
@@ -39,6 +50,38 @@ public Task<IEnumerable<InvitationTemplateModel>> GetAsync(string surveyId)
...
+                InvitationType = invitationTemplate.InvitationType
+            };
+
+            return Client.PatchAsJsonAsync(uri + "/" + invitationTemplate.Id, updatedInvitationTemplate)
+                .ContinueWith(responseMessageTask => responseMessageTask.Result.Content.ReadAsStringAsync().Result)
+                .ContinueWith(stringTask => JsonConvert.DeserializeObject<InvitationTemplateModel>(stringTask.Result))
+                .FlattenExceptions();
+        }
+
+        public Task<bool> RemoveAsync(string surveyId, string templateId)
```
maartje: InvitationTemplateModel.Id is an integer, so maybe we should make this templateId argument also an integer?

Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
```diff
@@ -39,6 +51,40 @@ public Task<IEnumerable<InvitationTemplateModel>> GetAsync(string surveyId)
...
+        public Task<bool> RemoveAsync(string surveyId, InvitationTemplateModel invitationTemplate)
+        {
+            CheckRequiredStringArgument(surveyId, nameof(surveyId));
+            CheckRequiredArgument(invitationTemplate, nameof(invitationTemplate));
+
+            var uri = SurveyInvitationTemplatesUrl(surveyId);
+
+            return Client.DeleteAsync(uri + "/" + invitationTemplate.Id)
+                .ContinueWith(response => response.Result.Content.ReadAsStringAsync().Result)
+                .ContinueWith(stringTask => JsonConvert.DeserializeObject<bool>(stringTask.Result))
```
marcoNipo: The delete from the api will return an object, you can check it on my PR 
https://github.com/NIPOSoftwareBV/nfield-source/pull/1100/files

Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
krakie: I replaced it with the invitationTemplate itself, because that's the thing you want to remove

Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
```diff
@@ -39,6 +51,40 @@ public Task<IEnumerable<InvitationTemplateModel>> GetAsync(string surveyId)
...
+                InvitationType = invitationTemplate.InvitationType
+            };
+
+            return Client.PatchAsJsonAsync(uri + "/" + invitationTemplate.Id, updatedInvitationTemplate)
+                .ContinueWith(responseMessageTask => responseMessageTask.Result.Content.ReadAsStringAsync().Result)
+                .ContinueWith(stringTask => JsonConvert.DeserializeObject<InvitationTemplateModel>(stringTask.Result))
+                .FlattenExceptions();
+        }
+
+        public Task<bool> RemoveAsync(string surveyId, InvitationTemplateModel invitationTemplate)
```
marcoNipo: Why the whole InvitationTemplateModel as parameter? you are using only the id

Library/Services/INfieldSurveyInvitationTemplatesService.cs
krakie: It should be an int, I've changed the SDK already

Library/Services/Implementation/NfieldSurveyInvitationTemplatesService.cs
krakie: updated the RemoveAsync method so expect an object

PR #31, opened 2017-08-09
Add images for invitation templates

marcoNipo:


Library/Services/Implementation/NfieldSurveyInvitationImagesService.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private string SurveyInvitationImagesTemplatesUrl(string surveyId, string filename)
+        {
+            var result = new StringBuilder(ConnectionClient.NfieldServerUri.AbsoluteUri);
+            result.AppendFormat(CultureInfo.InvariantCulture, @"Surveys/{0}/InvitationImages/{1}", surveyId, filename);
+            return result.ToString();
+        }
+
+        private static void CheckRequiredStringArgument(string argument, string name)
```
pilarodriguez: These methods (CheckRequiredStringArgument, CheckRequiredArgument) are duplicated in the code in a few places. It could be a good idea to move those methods to a helpers or another common class

PR #37, opened 2017-08-23
Implement domain email settings service

maartje:


Library/Models/DomainEmailSettingsEditable.cs
```diff
@@ -0,0 +1,25 @@
+namespace Nfield.Models
+{
+    public class DomainEmailSettingsEditable
```
marcoNipo: good idea, also SurveyEmailSettings could inherit from this

PR #44, opened 2017-08-30
[STORY] Yellow/ci update survey public ids

pilarodriguez:
New method to update the survey public ids. It will be used by the WD tests to avoid using the hidden api of the web manager

### DIFF
--- a/Library/Models/SurveyCounts.cs
+++ b/Library/Models/SurveyCounts.cs
@@ -1,7 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 
 namespace Nfield.Models
 {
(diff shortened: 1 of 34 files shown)


Library/Models/SurveyPublicId.cs
```diff
@@ -21,6 +21,11 @@ namespace Nfield.Models
     public class SurveyPublicId
     {
         /// <summary>
+        /// The public id that is used to link to the survey
+        /// </summary>
+        public string Id { get; set; }
```
neilboyd: I don't know why you would need this. The surveyId/linkType combination should be enough.

Library/Models/SurveyCounts.cs
```diff
@@ -32,6 +29,16 @@ public class SurveyCounts
...
 
         /// <summary>
+        /// Number of currently active live interviews
+        /// </summary>
+        public int? ActiveLiveCount { get; set; }
+
+        /// <summary>
+        /// Number of currently active test interviews
+        /// </summary>
+        public int? ActiveTestCount { get; set; }
```
neilboyd: Unrelated to this PR, but whatever...

PR #45, opened 2017-08-30
Include interview active counts properties to the response model of t…

pilarodriguez:
Include interview active counts properties to the response model of the survey counts


Library/Models/SurveyCounts.cs
```diff
@@ -32,6 +29,16 @@ public class SurveyCounts
         public int? RejectedCount { get; set; }
 
         /// <summary>
+        /// Number of currently active non-test interviews
```
carloscoceraperez: Replace "non-tests" with "live" so description is more direct and consistent with property name

PR #48, opened 2017-09-27
Reset sample polls activity  now

joukevandermaas:


Library/Services/Implementation/NfieldSurveySampleService.cs
```diff
@@ -103,10 +107,45 @@ public Task<int> ResetAsync(string surveyId, string respondentKey)
...
-                .ContinueWith(stringResult => JsonConvert.DeserializeObject<SampleResetStatus>(stringResult.Result).ResetCount)
+                .ContinueWith(stringResult => JsonConvert.DeserializeObject<BackgroundActivityStatus>(stringResult.Result).ActivityId)
+                .ContinueWith(activityResult => GetActivityResultAsync(activityResult.Result))
+                .Unwrap()
                 .FlattenExceptions();
         }
 
+        /// <summary>
+        /// Recursive method that polls the activity status until it completes.
+        /// </summary>
```
neilboyd: Wouldn't it be better to loop instead of recurse?

Library/Services/Implementation/NfieldSurveySampleService.cs
```diff
@@ -103,10 +107,45 @@ public Task<int> ResetAsync(string surveyId, string respondentKey)
...
+                        case 1: // started
+                            Thread.Sleep(millisecondsTimeout: 200);
+                            return GetActivityResultAsync(activityId);
+                        case 2: // succeeded
+                            var tcs = new TaskCompletionSource<int>();
+                            tcs.SetResult(obj["ResetTotal"].Value<int>());
+                            return tcs.Task;
+                        case 3: // failed
+                        default: 
+                            throw new Exception("Reset did not complete successfully");
```
neilboyd: Don't throw generic `Exception`.

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions
> Things to Avoid When Throwing Exceptions

> Do not throw Exception, SystemException, NullReferenceException, or IndexOutOfRangeException intentionally from your own source code.

Library/Services/Implementation/NfieldSurveySampleService.cs
joukevandermaas: I don't think it's possible without a synchronous `Wait`, or without increasing the .NET version and using async/await (which I don't think we should do as part of this story).

neilboyd: Maybe not possible in this one-liner, but I'm sure it's possible.
If someone uses this to wait for something that takes hours then it'll get a stack overflow pretty quickly. I assume.

joukevandermaas: As discussed in person, it turns out this doesn't actually increase the number of frames on the stack.

PR #51, opened 2017-10-17
[Story] Convert 'delete/block sample' to activity

mvdr:


Library/Services/Implementation/NfieldSurveySampleService.cs
```diff
@@ -64,15 +64,17 @@ public Task<int> DeleteAsync(string surveyId, string respondentKey)
...
 
-            var uri = SurveySampleUrl(surveyId);
+            var uri = SurveySampleUrl(surveyId) + @"/Delete";
             var filters = new List<SampleFilter>
             {
                 new SampleFilter{Name = "RespondentKey", Op = "eq", Value = respondentKey}
             };
 
-            return Client.DeleteAsJsonAsync<IEnumerable<SampleFilter>>(uri, filters)
+            return Client.PutAsJsonAsync<IEnumerable<SampleFilter>>(uri, filters)
```
marcoNipo: Are you sure it is a put now? I looked at the changes in the api and I couldn't see the change to make it a put anywhere... Am I missing something?

mvdr: You are right, in the API this should be a DELETE and not a PUT.

PR #56, opened 2017-12-05
CRUD script fragment

neilboyd:


neilboyd:
@joukevandermaas they're both CRUD, so 4 methods on each service

PR #58, opened 2017-12-06
implementation of script fragment service

neilboyd:


Library/Services/Implementation/NfieldSurveyScriptFragmentService.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            return Client.GetAsync(FragmentsApi(surveyId, null).AbsoluteUri)
+                .ContinueWith(responseMessageTask => responseMessageTask.Result.Content.ReadAsStringAsync().Result)
+                .ContinueWith(stringTask => JsonConvert.DeserializeObject<List<string>>(stringTask.Result).AsQueryable())
+                .FlattenExceptions();
+        }
+
+        public Task<string> GetAsync(string surveyId, string fileName)
+        {
+            if (string.IsNullOrEmpty(surveyId))
```
marcoNipo: you can use the Ensure class under the Utilities folder for these checks

Library/Services/Implementation/NfieldSurveyScriptFragmentService.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                throw new ArgumentNullException("surveyId");
+            }
+            if (string.IsNullOrEmpty(fileName))
+            {
+                throw new ArgumentNullException("fileName");
+            }
+
```
marcoNipo: Should we check the script as well here? Or should we leave it to the api?

PR #76, opened 2018-03-19
Implement Versioning NuGet package

leonardstronkhorst:
TODO:
- [x] determine if build is triggered by removing pre-release flag, and consequently use no suffix
https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables?tabs=batch


version.txt
```diff
@@ -0,0 +1,5 @@
+# Versioning of the NuGet package should be done in the following format: 
+# {major}.{minor}.{buildId}{suffix} where suffix should be either 
+# "-alpha", "-beta" or "" (empty) for respectively non-master-branches, master-branches
+# and release-versions (which is triggered once the prerelease tag is removed)
```
marcoNipo: It actually is when a release is published: the azure function that triggers the build does not check whether the prerelease **flag** is set

PR #93, opened 2018-04-13
Added SampleUpdate call for public SDK

JoseLuisNipo:
New call in public SDK for the "Sample update" feature

### DIFF
--- a/Library/Models/ClearSurveySampleModel.cs
+++ b/Library/Models/ClearSurveySampleModel.cs
@@ -18,7 +18,7 @@ using System.Collections.Generic;
 namespace Nfield.Models
 {
     /// <summary>
-    /// Model that holds the encrypted data and the initialization vector that was used to encrypt it
+    /// Model used to clear columns from a Sample Model
     /// </summary>
     public class ClearSurveySampleModel
     {
--- a/Library/Models/ClearSurveySampleModel.cs
+++ b/Library/Models/ClearSurveySampleModel.cs
@@ -13,36 +13,23 @@
 //    You should have received a copy of the GNU Lesser General Public License
 //    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
 
-using System;
+using System.Collections.Generic;
 
 namespace Nfield.Models
 {
     /// <summary>
-    /// class to define sample filters
+    /// Model that holds the encrypted data and the initialization vector that was used to encrypt it
     /// </summary>
-    public class SampleFilter
+    public class ClearSurveySampleModel
     {
         /// <summary>
-        /// Name of a field in a sample record
+        /// [MANDATORY] Filters to be applied for the clear operation
         /// </summary>
-        public string Name { get; set; }
-      
+        public IEnumerable<SampleFilter> Filters { get; set; }
+
         /// <summary>
-        /// Operator, supported operators:
-        /// "con" : contains
-        /// "ncon": not contains
-        /// "eq"  : equals
-        /// "lt"  : less than
-        /// "gt"  : greater than
-        /// "lte" : less or equal
-        /// "gte" : greater or equal
-        /// "neq" : not equal
+        /// The name of the columns to be cleared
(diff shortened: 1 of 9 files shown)


Library/Models/ClearSurveySampleModel.cs
```diff
@@ -0,0 +1,35 @@
...
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System.Collections.Generic;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Model that holds the encrypted data and the initialization vector that was used to encrypt it
```
evadelnuevo: Is this the correct description?

Library/Models/SampleColumnUpdate.cs
```diff
@@ -0,0 +1,35 @@
...
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Survey setting.
```
evadelnuevo: Is this the correct description? (also in ColumnName and Value)

Library/Models/SampleFilter.cs
```diff
@@ -20,7 +20,7 @@ namespace Nfield.Models
     /// <summary>
     /// class to define sample filters
     /// </summary>
-    internal class SampleFilter
+    public class SampleFilter
```
evadelnuevo: Is this change intended?

Tests/Services/NfieldSurveySampleTests.cs
```diff
@@ -527,6 +525,74 @@ public void TestClearAsync_ParamsAreOk_Successful()
...
+        {
+            int sampleRecordId = 1;
+            var columns = new List<SampleColumnUpdate>();
+            columns.Add(new SampleColumnUpdate { ColumnName = "dummyCol", Value = "dummyVal" });
+
+            var mockedNfieldConnection = new Mock<INfieldConnectionClient>();
+            var mockedHttpClient = CreateHttpClientMock(mockedNfieldConnection);
+
+            mockedHttpClient.Setup(client => client.PutAsJsonAsync($"{ServiceAddress}Surveys/{SurveyId}/Sample/Update",
+                It.Is<SurveyUpdateSampleRecordModel>(c => string.IsNullOrEmpty(SurveyId) && c.ColumnUpdates.Any(n => n.ColumnName == "dummyCol"))))
```
evadelnuevo: The SurveyId condition is unnecessary, instead, you should add the sampleRecordId to the condition, as it's used to build the SurveyUpdateSampleRecordModel

evadelnuevo:
Another thing: the PR should be against your ci branch (not master)

Library/Models/ClearSurveySampleModel.cs
JoseLuisNipo: no - fixed in commit 61b186c5103693627e2922122aa481b00ae36a20

Library/Models/SampleColumnUpdate.cs
JoseLuisNipo: no - fixed in commit 61b186c5103693627e2922122aa481b00ae36a20

Library/Models/SampleFilter.cs
JoseLuisNipo: yes, it's necessary

Tests/Services/NfieldSurveySampleTests.cs
JoseLuisNipo: fixed in commit 61b186c5103693627e2922122aa481b00ae36a20

Library/Models/SampleFilter.cs
evadelnuevo: 👍

PR #94, opened 2018-04-13
[STORY]Purple/ci update sampledata

JoseLuisNipo:
A new call ( `UpdateAsync` ) has been added to the SDK to update Sample data.

### DIFF
--- a/Library/Models/SampleColumnUpdate.cs
+++ b/Library/Models/SampleColumnUpdate.cs
@@ -18,30 +18,17 @@ using System;
 namespace Nfield.Models
 {
     /// <summary>
-    /// class to define sample filters
+    /// Model used to update columns from a Sample Model
     /// </summary>
-    public class SampleFilter
+    public class SampleColumnUpdate
     {
         /// <summary>
-        /// Name of a field in a sample record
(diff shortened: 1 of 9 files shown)


Library/Models/SurveyUpdateSampleRecordModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
```
hcsauniere: Initial comment missing:

//    This file is part of Nfield.SDK.
....

Library/Models/SampleColumnUpdate.cs
```diff
@@ -0,0 +1,35 @@
...
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Model used to update columns from a Sample Model
```
hcsauniere: It could be more clear to use _SurveyUpdateSampleRecordModel_ or _Sample Record_ instead of _Sample Model_

PR #100, opened 2018-04-16
search fields setting tests

DfernandezNipo:
and fix a wrong method name in the service

### DIFF
--- a/Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
+++ b/Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
@@ -40,8 +40,6 @@ namespace Nfield.Services
         [Fact]
         public void TestGetAsync_ReturnsData()
         {
-            
-
             var target = new NfieldDomainSearchFieldsSettingService();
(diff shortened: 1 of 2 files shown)


Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        #endregion
+
+        #region Put
+
+        [Fact]
+        public void TestPutAsync_SettingsNull_Throws()
+        {
+            var target = new NfieldDomainEmailSettingsService();
```
pilarodriguez: this should be NfieldDomainSearchFieldsSettingService, right?

Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
```diff
@@ -0,0 +1,85 @@
...
+                "field3"
+            }
+        };
+
+        #region Get
+
+        [Fact]
+        public void TestGetAsync_ReturnsData()
+        {
+
```
pilarodriguez: to many extra breaklines in this file

PR #103, opened 2018-04-23
Clear by interview id

pilarodriguez:
Clear also by interview id. Capi surveys don't have respondent key, so we should be also able to clear sample data by interview id

### DIFF
--- a/Library/Services/Implementation/NfieldSurveySampleService.cs
+++ b/Library/Services/Implementation/NfieldSurveySampleService.cs
@@ -133,22 +133,33 @@ namespace Nfield.Services.Implementation
                 .FlattenExceptions();
         }
 
-        public Task<int> ClearAsync(string surveyId, string respondentKey, IEnumerable<string> columnsToClear)
+        public Task<int> ClearAsync(string surveyId, string respondentKey, int? interviewId, IEnumerable<string> columnsToClear)
         {
             Ensure.ArgumentNotNullOrEmptyString(surveyId, nameof(surveyId));
-            Ensure.ArgumentNotNullOrEmptyString(respondentKey, nameof(respondentKey));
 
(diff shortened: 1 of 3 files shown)


Library/Services/Implementation/NfieldSurveySampleService.cs
```diff
@@ -133,22 +133,33 @@ public Task<int> ResetAsync(string surveyId, string respondentKey)
...
             var uri = SurveySampleUrl(surveyId) + @"/Clear";
 
-            var filters = new List<SampleFilter>
+            var filters = new List<SampleFilter>();
+
+            if (!string.IsNullOrEmpty(respondentKey))
             {
-                new SampleFilter{Name = "RespondentKey", Op = "eq", Value = respondentKey}
-            };
+                filters.Add(new SampleFilter{Name = "RespondentKey", Op = "eq", Value = respondentKey});
```
evadelnuevo: Maybe you could use the constants inside the SampleRecordColumn class for the names of the filters

Tests/Services/NfieldSurveySampleTests.cs
```diff
@@ -447,66 +452,107 @@ public void TestClearAsync_SurveyIdIsWhitespace_Throws()
...
             Assert.Throws<ArgumentException>(() =>
-                UnwrapAggregateException(target.ClearAsync(SurveyId, "anything", null)));
+                UnwrapAggregateException(target.ClearAsync(SurveyId, "anything", 1, new List<string>())));
         }
 
         [Fact]
-        public void TestClearAsync_ColumnsAreEmptyList_Throws()
+        public void TestClearAsync_ClearByRespondentKey_Successful()
         {
+            var mockedNfieldConnection = new Mock<INfieldConnectionClient>();
```
evadelnuevo: You can extract these as variables

PR #110, opened 2018-06-08
Blue/external api monitoring downloadlog

kefishev:


Library/Services/Implementation/NfieldExternalApisLogService.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        #endregion
+
+        private INfieldHttpClient Client
+        {
+            get { return ConnectionClient.Client; }
+        }
+
+        private Uri ExternalApiDownloaLogDatadUri
```
MartinRijks: The `d` travelled all the way to the `U`

PR #113, opened 2018-06-12
[Story] External APIs: Monitoring (users)

bjorncoltof:
This part of the feature covers adding functionality to the system that allows our users to monitor the performance of external API's as seen from our software.

Log request and response (verbatim)
Logging requests and purging on set time acceptable from GDPR ('x' days, setting)
Download request history per API through Nfield public API
Pass remaining interview metadata along on API calls (respondentkey?)
Design structure in blob in such a way that ops can easily download for domain/time-period

Related:
https://github.com/NIPOSoftwareBV/nfield-interviewengine/pull/83
https://github.com/NIPOSoftwareBV/Nfield-Manager-Surveys/pull/189
https://github.com/NIPOSoftwareBV/nfield-source/pull/2277
https://github.com/NIPOSoftware/Nfield-SDK/pull/113


Library/Services/Implementation/NfieldExternalApisLogService.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// Implementation of <see cref="INfieldExternalApisLogService"/>
+    /// </summary>
+    internal class NfieldExternalApisLogService : INfieldExternalApisLogService, INfieldConnectionClientObject
+    {
+        /// <summary>
+        /// See <see cref="INfieldExternalApisLogService.PostAsync"/>
+        /// </summary>
+        public Task<string> PostAsync(ExternalApiLogDownload logDownloadRequest)
+        {
+            if (logDownloadRequest == null)
```
marcoNipo: you could use `Ensure.ArgumentNotNull`

PR #117, opened 2018-09-21
Add collections with AAD Authentication

leonardstronkhorst:
This PR adds Postman Nfield collections for AAD: these are the same collections as for logging in the traditional way, except the collection with the SignIn request is removed (because we're not logging in via the SignIn request with AAD). Instead, you need to retrieve a bearer token via the ADTokenHelper and paste this into the bearer token field in Postman. In addition, the authorization type of the requests is Bearer Token (which automatically adds an authorization token to the header); and a key with 'X-Nfield-Domain' is added to the header. 

Retrieving a bearer token via OAuth 2.0 via the Postman 'Get New Access Token' form also works, but you need to enter URLs and a client id hardcoded. It would be really useful if you could use variables as in any other form in Postman, but this is not yet supported in the 'Get New Access Token' form (this feature is requested: https://github.com/postmanlabs/postman-app-support/issues/3525).


Postman/Aad/Aad-authentication.md
```diff
@@ -0,0 +1,21 @@
...
+This tool can be found
+[here](https://ktglbuc.sharepoint.com/sites/niposoftware/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2Fniposoftware%2FShared%20Documents%2FTools%20and%20Installers%2FADTokenHelper).
+You need to download all the files, or the program won't work. You can
+double-click the executable to get a token. This token will be copied to your
+clipboard. Note that tokens expire at some point. When that happens you can
+just double-click the executable again to get a new one.
+
+## Using Postman
+
+The collections in this folder use AAD Authentication. On the `Authorization` tab of each request, the authorization type is set to `Inherent auth from parent`. In the collection folder, click edit to edit the collection. On the `Authorization` tab, the option `Bearer Token` is selected. In the `Token` textbox, you should paste the token you got from the ADTokenHelper tool. You can now update and close the collection.
```
marcoNipo: A couple of screenshots would help here maybe

PR #122, opened 2018-09-25
[Story] Expand sdk with surveystate

hnuguse:
https://github.com/NIPOSoftwareBV/nfield/issues/1102


Library/Models/SurveyState.cs
```diff
@@ -20,7 +20,7 @@ namespace Nfield.Models
     /// <summary>
     /// The fieldwork status of the survey
     /// </summary>
-    public enum SurveyStatus
+    public enum SurveyState
```
joukevandermaas: You can't do this, this is a breaking change.

carloscoceraperez: What is it breaking? Its an internal model and all it's usages have been updated. At the end it's parsed into the same that SurveyStatus

carloscoceraperez: You mean someone which was expecting a Survey.Status in the response and now that code it broken?

joukevandermaas: it's public, not internal. so someone could have used it

PR #126, opened 2018-10-02
Add new client

joukevandermaas:
This PR moves the spot where the current token is kept to the NfieldConnection, while keeping the logic for the refresh in the client.

It also introduces a new client, to be used when doing Active Directory authentication.

### DIFF
--- a/Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
+++ b/Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
@@ -14,62 +14,29 @@
 //    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
 
 
-using Newtonsoft.Json;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using System.Text;
 using System.Threading.Tasks;
 
 namespace Nfield.Infrastructure
 {
-    internal sealed class ActiveDirectoryNfieldHttpClient : INfieldHttpClient
+    internal sealed class BearerTokenNfieldHttpClient : NfieldHttpClientBase
     {
-        private readonly NfieldConnection _connection;
+        private readonly string _domainName;
+        private readonly string _token;
 
-        public ActiveDirectoryNfieldHttpClient(NfieldConnection connection) => _connection = connection;
-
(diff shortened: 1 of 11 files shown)


Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using Newtonsoft.Json;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nfield.Infrastructure
+{
+    internal sealed class ActiveDirectoryNfieldHttpClient : INfieldHttpClient
```
LNieuwmegen: Maybe we should call this `BearerTokenHttpClient`?

Library/Infrastructure/DefaultNfieldHttpClient.cs
```diff
@@ -29,12 +29,16 @@ namespace Nfield.Infrastructure
     /// A wrapper around <see cref="HttpClient"/> that also adds an authentication header to the response
     /// if it was received in the request.
     /// </summary>
-    internal sealed class NfieldHttpClient : INfieldHttpClient
+    internal sealed class DefaultNfieldHttpClient : INfieldHttpClient
```
LNieuwmegen: `NfieldTokenHttpClient`?

Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            _domainName = domainName;
+            _token = token;
+        }
+
+        public override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
+        {
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", _token);
+            request.Headers.Add("X-Nfield-Domain", _domainName);
+
```
mvdr: Not for every call a domain header is needed (domainassignments endpoint), so probably check if _domainName has a value and based on that add a header or leave it out.

joukevandermaas: we don't (currently) have the ability to make domain assignments through the SDK, so for all currently supported endpoints it *is* needed.

if in the future this changes, let's address that problem then.

PR #131, opened 2018-10-10
[story] Add AAD authentication support

joukevandermaas:
This PR does the following high-level things:

1. Deprecate and remove the need for the dependency resolver for consumers of the package. What this basically means is that the "default" implementation can now also resolve services, which means by default people won't have to register one anymore.

2. Add a new way to sign in, which can be used when using AAD to authenticate. We did this by adding a new (derived) interface `INfieldConnectionV2` to keep backwards compatibility. The (internal) interface `INfieldHttpClient` now has two implementations, one for each authentication scheme.

3. Fix URI building. There were some instances where we generated paths like `/v1/surveys/id//addresses`. We made everything consistently use the `System.Uri` constructor instead. This is the bulk of the files changed, but hopefully fairly straightforward to review.


Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
```diff
@@ -0,0 +1,62 @@
...
+            try
+            {
+                return response.ValidateResponse();
+            }
+            catch (NfieldHttpResponseException ex)
+            {
+                if (ex.Message.Contains("X-NFIELD-DOMAIN"))
+                {
+                    // message is not very useful in the SDK case
+                    throw new NfieldHttpResponseException(ex.HttpStatusCode, $"Invalid domain '{_domainName}'");
```
neilboyd: Should we be doing `Invariant` everywhere in this project?

Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            var response = await Client.SendAsync(request);
+
+            try
+            {
+                return response.ValidateResponse();
+            }
+            catch (NfieldHttpResponseException ex)
+            {
+                if (ex.Message.Contains("X-NFIELD-DOMAIN"))
```
neilboyd: Why is this uppercase, when a few lines above it was mixed case?

Library/Extensions/NfieldConnectionClientExtensions.cs
```diff
@@ -29,12 +30,10 @@ namespace Nfield.Extensions
...
     public static class NfieldConnectionClientExtensions
     {
-        private static string BackgroundActivityUrl(this INfieldConnectionClient client, string activityId)
+        private static Uri BackgroundActivityUrl(this INfieldConnectionClient client, string activityId)
         {
-            var result = new StringBuilder(client.NfieldServerUri.AbsoluteUri);
-            result.AppendFormat(CultureInfo.InvariantCulture, @"BackgroundActivities/{0}", activityId);
-
-            return result.ToString();
+            return new Uri(client.NfieldServerUri, string.Format(CultureInfo.InvariantCulture,
```
neilboyd: Can we start using string interpolation when we refactor code?

Library/Infrastructure/DefaultNfieldHttpClient.cs
```diff
@@ -0,0 +1,52 @@
...
+    internal sealed class DefaultNfieldHttpClient : NfieldHttpClientBase
+    {
+        private string _token;
+
+        public DefaultNfieldHttpClient(HttpClient client) : base(client)
+        { }
+
+        public override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request)
+        {
+            request.Headers.Authorization = new AuthenticationHeaderValue("Basic", _token);
```
neilboyd: does this mean the value is null the first time we send it?

Library/Infrastructure/NfieldConnection.cs
```diff
@@ -67,27 +72,36 @@ public Task<bool> SignInAsync(string domainName, string username, string passwor
...
                     var result = responseMessageTask.Result;
-
-                    _token = Client.AuthToken;
-                    
                     return result.StatusCode == HttpStatusCode.OK;
-
                 }).FlattenExceptions();
         }
 
+        public Task SignInAsync(string domainName, Func<Task<string>> provideTokenAsync)
```
neilboyd: There's no comment. Why doesn't this method do anything?

Library/Services/Implementation/NfieldDomainEmailSettingsService.cs
```diff
@@ -55,11 +56,9 @@ public void InitializeNfieldConnection(INfieldConnectionClient connection)
             ConnectionClient = connection;
         }
 
-        private string DomainEmailSettingsUrl()
+        private Uri DomainEmailSettingsUrl()
         {
-            var result = new StringBuilder(ConnectionClient.NfieldServerUri.AbsoluteUri);
-            result.AppendFormat(CultureInfo.InvariantCulture, @"EmailSettings");
```
neilboyd: weird that it was doing `AppendFormat` when there was no format :-/

Library/Services/Implementation/NfieldInterviewsService.cs
```diff
@@ -66,11 +67,11 @@ private static void CheckSurveyId(string surveyId)
...
         private INfieldHttpClient Client => ConnectionClient.Client;
 
-        private string InterviewsApiUri(string surveyId, int interviewId)
+        private Uri InterviewsApiUri(string surveyId, int interviewId)
         {
-            var uriText = new StringBuilder(ConnectionClient.NfieldServerUri.AbsoluteUri);
-            uriText.Append($"Surveys/{surveyId}/Interviews/{interviewId}");
-            return new Uri(uriText.ToString()).AbsoluteUri;
+            return new Uri(ConnectionClient.NfieldServerUri,
+                string.Format(CultureInfo.InvariantCulture, "Surveys/{0}/Interviews/{1}",
```
neilboyd: Why not keep it as string interpolation? Consistency with the code above here?

Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
```diff
@@ -46,7 +46,7 @@ public NfieldDomainSearchFieldsSettingServiceTests()
         [Fact]
         public void TestGetAsync_ReturnsData()
         {
-            var mockClient = InitMockClientGet(ServiceAddress + @"SearchFieldsSetting", _expected);
+            var mockClient = InitMockClientGet(new Uri(ServiceAddress, @"SearchFieldsSetting"), _expected);
```
neilboyd: Why `@`?

Tests/Services/NfieldDomainSearchFieldsSettingServiceTests.cs
```diff
@@ -67,7 +67,7 @@ public void TestPutAsync_SettingsNull_Throws()
         [Fact]
         public void TestPutAsync_ReturnsData()
         {
-            var mockClient = InitMockClientPut(ServiceAddress + @"SearchFieldsSetting", _expected, _expected);
+            var mockClient = InitMockClientPut(new Uri(ServiceAddress, @"SearchFieldsSetting"), _expected, _expected);
```
neilboyd: ditto

Library/Infrastructure/NfieldConnection.cs
neilboyd: The comment on the interface is helpful, but it says "sign in" but it doesn't appear to actually do that.

Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
joukevandermaas: It's upper case in the error message. The case is ignored when provided in the headers of a request.

Library/Infrastructure/DefaultNfieldHttpClient.cs
joukevandermaas: I suppose it does.. it was already this way (just in another file) and I didn't look too closely at it. Apparently the API ignores this header on the `SignIn` request.

Library/Infrastructure/NfieldConnection.cs
joukevandermaas: We were struggling to find a better name :smile: if you have a suggestion we can change it.

Library/Services/Implementation/NfieldDomainEmailSettingsService.cs
joukevandermaas: copy/paste from somewhere else i assume.

Library/Infrastructure/ActiveDirectoryNfieldHttpClient.cs
joukevandermaas: invariant on string interpolation requires an extra nuget reference. i'd rather not add that if not super required.

Library/Extensions/NfieldConnectionClientExtensions.cs
joukevandermaas: not in the general case, unless we add a nuget package to make those invariant culture.

Library/Infrastructure/NfieldConnection.cs
neilboyd: How about `RegisterTokenProvider`? It also doesn't need to be async.

PR #133, opened 2018-10-15
delete an unused file

neilboyd:


bjorncoltof:
I know this counts towards hacktober, but a pr that adds strong naming back in is what we really need. The removal of signing in March should not have happened. Not signing is OK for internal stuff, public packages should simply be signed

bjorncoltof:
Only meant to comment

neilboyd:
I created PR #136 to sign it again.
But let's keep this one open for now in case this turns into a discussion about whether to sign it or not.
And in any case. we could still delete `packages/repositories.config`.

PR #135, opened 2018-10-15
use string interpolation

neilboyd:


neilboyd:
@bjorncoltof I replaced all occurrences of `string.Format`

PR #82, opened 2018-03-27
White/ci update development

marcoNipo:
Bringing the development branch up to speed with master in order to cleanly get rid of it, as part of [the new release procedure](https://github.com/NIPOSoftwareBV/Nfield-Documentation/pull/319)

PR #136, opened 2018-10-15
Sign the assembly again.

neilboyd:
It shouldn't have been unsigned in PR #82.


bjorncoltof:
See https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries

PR #142, opened 2018-11-13
Purple/Update nipo .net standard packages 

evadelnuevo:
Update the packages we changed to .NET standard and update this one to .NET Standard:
- Nfield.Quota: https://github.com/NIPOSoftware/Nfield.Quota/pull/45


evadelnuevo:
@joukevandermaas You're right! I'll change it to 2.0
cc @pilarodriguez

bjorncoltof:
Why not upgrade SDK to .Net standard 2.0?

evadelnuevo:
@bjorncoltof We were going to do it in a different story later, but we could probably do it now.

So we will try to do it now 😄

PR #154, opened 2019-03-08
Post survey sample data as JSON

basbakkernipo:
This PR will add the ability to send survey sample data as JSON, as well as CSV.

### DIFF
--- a/Library/Services/Implementation/NfieldSurveySampleService.cs
+++ b/Library/Services/Implementation/NfieldSurveySampleService.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Text;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Nfield.Extensions;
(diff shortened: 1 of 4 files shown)


Library/Services/Implementation/NfieldSurveySampleService.cs
```diff
@@ -55,6 +56,18 @@ public Task<SampleUploadStatus> PostAsync(string surveyId, string sample)
                 .ContinueWith(stringResult => JsonConvert.DeserializeObject<SampleUploadStatus>(stringResult.Result))
                 .FlattenExceptions();
         }
+        
+        public Task<SampleUploadStatus> PostJsonAsync<TContent>(string surveyId, TContent sample)
```
marcoNipo: You should add at least one [unit test](https://github.com/NIPOSoftware/Nfield-SDK/blob/master/Tests/Services/NfieldSurveySampleTests.cs) for this

basbakkernipo:
Decided not to implement JSON parser after all. We updated the CSV parser to check for quotes, instead of just splitting on delimiters.

See [this PR](https://github.com/NIPOSoftwareBV/Nfield-Manager-SurveysAndSampleManagement/pull/447) for the actual solution.

PR #156, opened 2019-03-22
add sdk version to http request

neilboyd:
https://github.com/NIPOSoftwareBV/nfield/issues/384


neilboyd:
The version is always 1.0.0 so this is not useful. It would require changes to the build pipeline in order to make this work.

PR #169, opened 2019-07-12
Add methods for listing survey groups and members

joukevandermaas:


Library/Services/Implementation/NfieldSurveyGroupService.cs
```diff
@@ -0,0 +1,73 @@
+﻿//    This file is part of Nfield.SDK.
```
LNieuwmegen: This shouldnt be on the other classes?

PR #172, opened 2019-07-17
[Story] List + CRUD on Survey Groups

mvdr:
[AB#53716](https://niposoftware.visualstudio.com/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/53716)


Tests/Services/NfieldSurveyGroupServiceTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            Assert.Equal(2, actualSurveyGroups.Count());
+
+            var defaultGroup = actualSurveyGroups.Single(sg => sg.SurveyGroupId == 1);
+            Assert.Equal("Default", defaultGroup.Name);
+            Assert.Equal(new DateTime(1799, 11, 10), defaultGroup.CreationDate);
+
+            var secondGroup = actualSurveyGroups.Single(sg => sg.SurveyGroupId == 2);
+            Assert.Equal("War & Peace", secondGroup.Name);
+            Assert.Equal("Napoleon satisfaction surveys", secondGroup.Description);
+            Assert.Equal(new DateTime(1812, 06, 26), secondGroup.CreationDate);
```
carloscoceraperez: Ditto

```suggestion
           Assert.Equal(expectedGroups.Where(g => g.SurveyGroupId == 1), new DateTime(1812, 06, 26), secondGroup.CreationDate);
```

Tests/Services/NfieldSurveyGroupServiceTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            var actualSurveyGroups = await target.GetAllAsync();
+
+            Assert.Equal(2, actualSurveyGroups.Count());
+
+            var defaultGroup = actualSurveyGroups.Single(sg => sg.SurveyGroupId == 1);
+            Assert.Equal("Default", defaultGroup.Name);
+            Assert.Equal(new DateTime(1799, 11, 10), defaultGroup.CreationDate);
+
+            var secondGroup = actualSurveyGroups.Single(sg => sg.SurveyGroupId == 2);
+            Assert.Equal("War & Peace", secondGroup.Name);
```
carloscoceraperez: Ditto
```suggestion
            Assert.Equal(expectedGroups.Where(g => g.SurveyGroupId == 2).Name ,  secondGroup.Name);
```

you do this similar logic using an array access instead of a LinQ expression, below

``` 
Assert.Equal(expectedSurveys[0].SurveyId, actualSurveys.ToArray()[0].SurveyId);
```

Check other asserts.

Tests/Services/NfieldSurveyGroupServiceTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            Assert.Equal(2, result.SurveyGroupId);
+
+            // modify survey group object we got back, then post it
+            result.Name = "Default with a twist";
+            result.Description = "Modified description";
+
+            result = await target.UpdateAsync(result.SurveyGroupId, result);
+
+            Assert.NotNull(result);
+            Assert.Equal("Default with a twist", result.Name);
```
carloscoceraperez: If you expect the _result.Name_ to be the same as second group SurveyGroup Name, you can use that object property value instead of a literal, e.g

```suggestion
           Assert.Equal(expectedGroups.Where(g => g.SurveyGroupId == 2).Name ,  result.Name);
```

This would make the test more robust IMO, as you base your assert condition on a test arrange value, not input literal

carloscoceraperez:
I was wondering why not the .ReturnsAsync() instead of that CreateTask until I opened the code 😄
Btw very clean CRUD!

Tests/Services/NfieldSurveyGroupServiceTests.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            Assert.NotNull(result);
+            Assert.Equal("Default", result.Name);
+            Assert.Equal(null, result.Description);
+            Assert.Equal(new DateTime(1799, 11, 10), result.CreationDate);
+            Assert.Equal(1, result.SurveyGroupId);
+        }
+
+        [Fact]
+        public async Task CanCreateModifyAndDeleteGroup()
```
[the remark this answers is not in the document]
mvdr: This test is meant for testing the whole scenario: create survey group, update it and then delete it.
So we leave it one test.

Library/Services/Implementation/NfieldSurveyGroupService.cs
```diff
@@ -0,0 +1,112 @@
...
+                return result;
+            }
+        }
+
+        public async Task DeleteAsync(int surveyGroupId)
+        {
+            var uri = new Uri(ConnectionClient.NfieldServerUri, $"SurveyGroups/{surveyGroupId}");
+
+            
+            using(await ConnectionClient.Client.DeleteAsync(uri))
```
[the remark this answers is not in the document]
mvdr: It is actually needed. DeleteAsync returns a HttpResponseMessage, which is IDisposable, so it should be disposed.

neilboyd:
This is in the public repo, so remove all the links to private repos in the description.

PR #174, opened 2019-07-24
Add method to move survey

joukevandermaas:


joukevandermaas:
@mvdr what kind of meaningful unit test can i write for this?

mvdr:
I agree that a test won't add much, because we now also have the API automated test. It is just that we have tests for all the other calls. But I leave it up to you. If you don't think it is needed, I am fine with that as well.

PR #177, opened 2019-08-21
[Story] CRUD for users

joukevandermaas:
AB#53833

### DIFF
--- a/Library/Models/LocalUser.cs
+++ b/Library/Models/LocalUser.cs
@@ -25,31 +25,26 @@ namespace Nfield.Models
         /// <summary>
         /// Name of user
         /// </summary>
-        [JsonProperty]
         public string UserName { get; set; }
 
         /// <summary>
         /// First name of user
         /// </summary>
-        [JsonProperty]
         public string FirstName { get; set; }
 
         /// <summary>
         /// Last name of user
         /// </summary>
-        [JsonProperty]
         public string LastName { get; set; }
 
         /// <summary>
         /// Email of the user
         /// </summary>
-        [JsonProperty]
         public string Email { get; set; }
 
         /// <summary>
         /// Role of the user
         /// </summary>
-        [JsonProperty]
         public string UserRole { get; set; }
     }
 
@@ -61,7 +56,6 @@ namespace Nfield.Models
         /// <summary>
         /// Password of the new user
         /// </summary>
-        [JsonProperty]
         public string Password { get; set; }
     }
 
--- a/Library/Models/LocalUser.cs
+++ b/Library/Models/LocalUser.cs
@@ -1,0 +1,79 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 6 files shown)


Library/Models/LocalUser.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    /// <summary>
+    /// Model representing the settable values in on a local user.
+    /// </summary>
+    public class LocalUserValues
+    {
+        /// <summary>
+        /// Name of user
+        /// </summary>
+        [JsonProperty]
```
bjorncoltof: Do we also need these on all the public properties?

joukevandermaas: I guess we probably don't? I will try remove them and see what happens.

Library/Models/LocalUser.cs
```diff
@@ -0,0 +1,73 @@
...
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Model representing the settable values in on a local user.
```
marcoNipo: "_in_ on a local user" :thinking:

Library/Models/LocalUser.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// Model representing a local user.
+    /// </summary>
+    public class LocalUser : LocalUserValues
+    {
+        /// <summary>
+        /// The Id of the local user.
+        /// </summary>
+        [JsonProperty]
```
marcoNipo: Do you need this attribute?

Library/Services/INfieldLocalUserService.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    /// <summary>
+    /// Service used to list, update, create and delete local users.
+    /// </summary>
+    public interface INfieldLocalUserService
+    {
+        /// <summary>
+        /// Retrieve a specific local user
+        /// </summary>
+        /// <param name="identityId"></param>
```
marcoNipo: If the comment is empty, you can delete the param node

Library/Services/Implementation/NfieldLocalUserService.cs
```diff
@@ -0,0 +1,107 @@
...
+        public async Task<LocalUser> CreateAsync(NewLocalUser model)
+        {
+            if (model == null)
+                throw new ArgumentNullException(nameof(model));
+
+            var uri = new Uri(ConnectionClient.NfieldServerUri, "LocalUsers");
+
+            using (var response = await ConnectionClient.Client.PostAsJsonAsync(uri, model))
+            {
+                var result = await DeserializeJsonAsync<LocalUser>(response);
```
marcoNipo: you don't really need the variable

Library/Services/Implementation/NfieldLocalUserService.cs
```diff
@@ -0,0 +1,107 @@
...
+        public async Task<LocalUser> UpdateAsync(string identityId, LocalUserValues model)
+        {
+            if (model == null)
+                throw new ArgumentNullException(nameof(model));
+
+            var uri = new Uri(ConnectionClient.NfieldServerUri, $"LocalUsers/{identityId}");
+
+            using (var response = await ConnectionClient.Client.PatchAsJsonAsync(uri, model))
+            {
+                var result = await DeserializeJsonAsync<LocalUser>(response);
```
marcoNipo: same here

Library/Services/Implementation/NfieldLocalUserService.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        public async Task DeleteAsync(string identityId)
+        {
+            var uri = new Uri(ConnectionClient.NfieldServerUri, $"LocalUsers/{identityId}");
+
+            // note: we need to dispose the response even when we don't use it
+            using (await ConnectionClient.Client.DeleteAsync(uri))
+            {
```
marcoNipo: Is it possible to do it without the brackets {}? Maybe with a simple `;` after the using?

bjorncoltof: .Dispose()?

PR #178, opened 2019-08-23
SurveyGroup assignments in SDK

mvdr:


Library/Models/AadObjectType.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Azure Active Directory object type
+    /// </summary>
+    public enum AadObjectType
+    {
+        /// <summary>
+        /// Unknown
+        /// </summary>
+        Unknown = 0,
```
joukevandermaas: I think this one shouldn't be in the sdk

Library/Models/NativeIdentityModel.cs
```diff
@@ -0,0 +1,28 @@
...
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace Nfield.SDK.Models
+{
+    /// <summary>
+    /// Model for native identity
+    /// </summary>
+    public class NativeIdentityModel
```
joukevandermaas: I think we can just ask for a string instead of wrapping it in a model.

PR #181, opened 2019-08-29
[Story] Endpoints to assign/unassign identity

mvdr:
Pull request for [PBI 53834](https://niposoftware.visualstudio.com/DefaultCollection/Nfield/_workitems/edit/53834)

SDK updated for new API endpoints: assignments to survey groups

AB#53834

### DIFF
--- a/Library/Services/Implementation/NfieldSurveyGroupAssignmentsService.cs
+++ b/Library/Services/Implementation/NfieldSurveyGroupAssignmentsService.cs
@@ -1,0 +1,109 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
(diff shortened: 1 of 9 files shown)


Library/Services/Implementation/NfieldSurveyGroupAssignmentsService.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Nfield.Infrastructure;
+using Nfield.Models;
+
+namespace Nfield.Services.Implementation
+{
+    internal class NfieldSurveyGroupAssignmentsService : INfieldSurveyGroupAssignmentsService, INfieldConnectionClientObject
+    {
+        public INfieldConnectionClient ConnectionClient { get; internal set; }
```
carloscoceraperez: Do you actually need to make this member internal if you have a public method to set it? I think you can make it private and just use the _Initialize_ to set it.

mvdr: Maybe you are right but all other existing service classes are internal as well, so I don't want to deviate from that.

PR #187, opened 2019-10-17
[Story] GH#1531: Public SDK: Allow unfixed positions during upload questionnaire

kefishev:
We added a property to the SurveyScript model to specify if script with unfixed position is allowed

### DIFF
--- a/Library/Models/SurveyScript.cs
+++ b/Library/Models/SurveyScript.cs
@@ -37,5 +37,10 @@ namespace Nfield.Models
         /// Warning messages that resulted from the parse
         /// </summary>
         public IEnumerable<string> WarningMessages { get; set; }
+
+        /// <summary>
+        /// Indicates if script with unfixed positions is allowed
+        /// </summary>
+        public bool UnfixedIsOk { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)


Library/Models/SurveyScript.cs
```diff
@@ -37,5 +37,10 @@ public class SurveyScript
         /// Warning messages that resulted from the parse
         /// </summary>
         public IEnumerable<string> WarningMessages { get; set; }
+
+        /// <summary>
+        /// Indicates if script with unfixed positions is allowed
+        /// </summary>
+        public bool UnfixedIsOk { get; set; }
```
marcoNipo: No new/updated unit tests for this?

kefishev: We added two integration test for this, Not sure what should I test in unit tests

marcoNipo: I was thinking about [these](https://github.com/NIPOSoftware/Nfield-SDK/blob/master/Tests/Services/NfieldSurveyScriptServiceTests.cs) but the api integration tests are way more useful of course

PR #189, opened 2020-01-08
Issue [#1834]: Extend API with calls for Management Tab

dquilong:
AB#60267

[Issue #1834](https://github.com/NIPOSoftwareBV/nfield/issues/1834)

New controller in the public API for retrieving all the survey resources has been created, this controller supports Odata filtering.

### DIFF
--- a/Library/Models/SurveyResources.cs
+++ b/Library/Models/SurveyResources.cs
@@ -1,0 +1,39 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Holds the properties of survey resources
+    /// </summary>
+    public class SurveyResources : SurveyBase
+    {
+        public DateTime? CreationDate { get; internal set; }
+
+        public string ClientName { get; internal set; }
+
+        [JsonProperty]
+        public string Owner { get; internal set; }
+
+        public DateTime? LastDataDownloadDate { get; internal set; }
+
+        public DateTime? LastDataCollectionDate { get; internal set; }
+
+        public long? Size { get; internal set; }
+    }
+}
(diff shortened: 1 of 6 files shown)


Library/Models/SurveyResources.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// Holds the properties of survey resources
+    /// </summary>
+    public class SurveyResources : SurveyBase
+    {
+        public DateTime? CreationDate { get; internal set; }
+
+        public string ClientName { get; internal set; }
+
+        [JsonProperty]
+        public string Owner { get; internal set; }
```
evadelnuevo: Why is this marked with a `JsonProperty` attribute?

Tests/Services/NfieldSurveyResourcesServiceTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                .Returns(CreateTask(HttpStatusCode.OK, new StringContent(JsonConvert.SerializeObject(expectedSurveyResources))));
+
+            var target = new NfieldSurveyResourcesService();
+            target.InitializeNfieldConnection(mockedNfieldConnection.Object);
+
+            var actualSurveyResources = target.QueryAsync().Result;
+
+            Assert.Equal(expectedSurveyResources[0].Owner, actualSurveyResources.ToArray()[0].Owner);
+            Assert.Equal(expectedSurveyResources[1].Owner, actualSurveyResources.ToArray()[1].Owner);
+            Assert.Equal(expectedSurveyResources[2].Owner, actualSurveyResources.ToArray()[2].Owner);
```
evadelnuevo: You should also assert: name, channel, creationdate, clientname and state

Tests/Services/NfieldSurveyResourcesServiceTests.cs
```diff
@@ -69,21 +69,24 @@ public void TestQueryAsync_ServerReturnsQuery_ReturnsListWithSurveyResources()
...
-                Assert.Equal(expectedSurveyResources[i].SurveyId, actualSurveyResources.ToArray()[i].SurveyId);
-                Assert.Equal(expectedSurveyResources[i].SurveyName, actualSurveyResources.ToArray()[i].SurveyName);
-                Assert.Equal(expectedSurveyResources[i].Owner, actualSurveyResources.ToArray()[i].Owner);
-                Assert.Equal(expectedSurveyResources[i].Size, actualSurveyResources.ToArray()[i].Size);
-                Assert.Equal(expectedSurveyResources[i].CreationDate, actualSurveyResources.ToArray()[i].CreationDate);
-                Assert.Equal(expectedSurveyResources[i].LastDataDownloadDate, actualSurveyResources.ToArray()[i].LastDataDownloadDate);
-                Assert.Equal(expectedSurveyResources[i].LastDataCollectionDate, actualSurveyResources.ToArray()[i].LastDataCollectionDate);
-                Assert.Equal(expectedSurveyResources[i].State, actualSurveyResources.ToArray()[i].State);
-                Assert.Equal(expectedSurveyResources[i].Channel, actualSurveyResources.ToArray()[i].Channel);
+                var actualSurveyResource = actualSurveyResources.ToArray()[i];
```
evadelnuevo: Maybe you can do actualSurveyResources.ElementAt(i). Now you are converting it to an array for each iteration of the loop

Library/Models/SurveyResources.cs
evadelnuevo: 🤔

dquilong: This is needed for the tests as they have a internal set, the Json deserializer can't create a new  SurveyResources object. This is somethin that we already do in the SurveyBase for the surveyName.

evadelnuevo: And the other properties are different? (ClientName, etc)

dquilong: all propperties have [JsonProperty] now, because all of them are being tested

PR #190, opened 2020-01-10
Issue [#1834]: Extend API with calls for Management Tab

dquilong:
[AB#60267](https://niposoftware.visualstudio.com/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/60267)

[Issue #1834](https://github.com/NIPOSoftwareBV/nfield/issues/1834)

New controller in the public API for retrieving all the survey resources has been created, this controller supports Odata filtering.

### DIFF
--- a/Library/Models/SurveyChannel.cs
+++ b/Library/Models/SurveyChannel.cs
@@ -1,0 +1,31 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Survey channel
+    /// </summary>
+    public enum SurveyChannel
+    {
+        Unknown,
+        Cati,
+        Online,
+        Capi,
+    }
+}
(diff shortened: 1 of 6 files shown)


Library/Models/SurveyChannel.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Survey channel
+    /// </summary>
+    public enum SurveyChannel
```
joukevandermaas: Incredible that this did not exist yet.. I guess because it was encoded in the survey type?

Library/Models/SurveyResources.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Holds the properties of survey resources
+    /// </summary>
+    public class SurveyResources : SurveyBase
```
joukevandermaas:
```suggestion
    public class SurveyResource : SurveyBase
```

Library/Services/Implementation/NfieldSurveyResourcesService.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        public Task<IQueryable<SurveyResources>> QueryAsync()
+        {
+            return Client.GetAsync(SurveyResourcesApi)
+                         .ContinueWith(
+                             responseMessageTask => responseMessageTask.Result.Content.ReadAsStringAsync().Result)
+                         .ContinueWith(
+                             stringTask =>
+                             JsonConvert.DeserializeObject<List<SurveyResources>>(stringTask.Result).AsQueryable())
+                         .FlattenExceptions();
```
joukevandermaas: You say "supports odata filtering" but that is not what is happening here; all resources are always fetched and the filter is applied client-side.

Library/Models/SurveyChannel.cs
dquilong: The SurveyType enum has this types: Basic, Advanced, EuroBarometer, OnlineBasic, EuroBarometerAdvanced and Cati. I guess it's not the same, I took glu-api as reference, so I copied this enum from there

joukevandermaas: Yeah, so CAPI has 4 survey types, online and CATI both have only one. This enum is more clear for sure.

Library/Services/Implementation/NfieldSurveyResourcesService.cs
dquilong: Yes, I added the odata filtering just to the controller in the API. It wasn't a requirement but It was easy and fast to add it so I did it.  Filtering could be used here as well, i'm going to check what I need to add it.

joukevandermaas: Don't worry about it too much, it doesn't seem like a huge deal. The surveys endpoint works the same way.

PR #200, opened 2020-05-08
[Story] Extend download (filter by Etag and include VAR file) (#66561)

dquilong:
[AB#66561](https://niposoftware.visualstudio.com/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/66561)

This is part of the feature [Download data for a specific version of the survey](https://github.com/NIPOSoftwareBV/Nfield-Documentation/blob/master/features/Download-Data-For-Specific-Survey-Version/grooming-notes.md)

- Update SurveyDownloadDataRequest model
- Update example, filtering sample data download with the SurveyVersion

### DIFF
--- a/Examples/Program.cs
+++ b/Examples/Program.cs
@@ -172,11 +172,13 @@ namespace Nfield.SDK.Samples
                 DownloadSuspendedLiveInterviewData = false,
                 DownloadCapturedMedia = false,
                 DownloadParaData = false,
+                DownloadVarFile = false,
                 DownloadTestInterviewData = true,
                 DownloadFileName = "MyFileName",
                 StartDate = DateTime.Today.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture), // UTC time start of today
                 EndDate = DateTime.Today.AddDays(1).ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture), // UTC time end of today
-                SurveyId = "SomeSurveyId"
+                SurveyId = "SomeSurveyId",
+                SurveyVersion = 637242848690284790 // If the SurveyVersion (Etag) is specified only the surveys matching this version will be downloaded
             };
 
             var task = surveyDataService.PostAsync(myRequest).Result;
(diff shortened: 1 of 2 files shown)


Library/Models/SurveyDownloadDataRequest.cs
```diff
@@ -94,5 +99,11 @@ public class SurveyDownloadDataRequest
         /// May be null or empty if no end date is specified.
         /// </summary>
         public string EndDate { get; set; }
+
+        /// <summary>
+        /// The survey version (Etag)
+        /// May be null or 0 if no survey version is specified.
+        /// </summary>
+        public long? SurveyVersion { get; set; }
```
evadelnuevo, line 107: Why is this a long instead of a string?

hnuguse: It is the the etag, which is a number

Examples/Program.cs
```diff
@@ -172,11 +172,13 @@ static void Main()
...
                 DownloadCapturedMedia = false,
                 DownloadParaData = false,
+                DownloadVarFile = false,
                 DownloadTestInterviewData = true,
                 DownloadFileName = "MyFileName",
                 StartDate = DateTime.Today.ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture), // UTC time start of today
                 EndDate = DateTime.Today.AddDays(1).ToUniversalTime().ToString("yyyy-MM-ddTHH:mm:ssZ", CultureInfo.InvariantCulture), // UTC time end of today
-                SurveyId = "SomeSurveyId"
+                SurveyId = "SomeSurveyId",
+                SurveyVersion = 637242848690284790 // If the SurveyVersion (Etag) is specified only the surveys matching this version will be downloaded
```
hnuguse, line 181: Can't you keep the variable name as Etag? Or is that too Techhie? This also saves you the hassle of mapping and such on the api side..

dquilong: We decided internally we will call it Etag but externally Survey version. In the manager it will appear as survey version. See example:

https://xd.adobe.com/view/7095f99c-adc2-4cb7-4878-2de7845ea13f-06a4/

Library/Models/SurveyDownloadDataRequest.cs
dquilong: In the database the Etag is a long int

pilarodriguez: I think the endpoint Carlos is creating (to retreive the list of etag, is returning an string (converting the long to string). Not sure which option is better, but both should have the same type (both long or both string)

PR #201, opened 2020-05-11
[Story] Extend download (filter by Etag and include VAR file) (#66561)

dquilong:
AB#66561

This is part of the feature Download data for a specific version of the survey

Update SurveyDownloadDataRequest model
Update example, filtering sample data download with the SurveyVersion


dquilong:
Yes, we had in mind to update the version in the other story. Both should go in this sprint. Thanks fo the reminder :)

PR #206, opened 2020-07-14
add quota service

behnazbeaumont:


Library/Models/QuotaFrameVersion.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// The unique identifier for a quota frame version
+        /// </summary>
+        public string Id { get; set; }
+
+        /// <summary>
+        /// The version of the quota frame
+        /// </summary>
+        [JsonProperty(PropertyName = "eTag")]
+        public string Etag { get; set; }
```
joukevandermaas, line 35: Let's check if we use the word elsewhere in this project and stick to consistent capitalization.

Library/Services/Implementation/NfieldQuotaService.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        private Uri QuotaFrameVersionsApi(string surveyId)
+        {
+            return new Uri(ConnectionClient.NfieldServerUri, $"Surveys/{surveyId}/QuotaVersions");
+        }
+
+        private static void ValidateParams(string surveyId)
+        {
+            if (surveyId == null)
```
joukevandermaas, line 51: `string.IsNullOrEmpty`? I guess you can't use that because you want to throw different exceptions?

Library/Models/QuotaFrameVersion.cs
behnazbeaumont: used this for consistency with survey version, can remove it if its not needed.

Library/Services/Implementation/NfieldQuotaService.cs
behnazbeaumont: it was done this way in the survey version service so for consistency I did it the same way, can change it if you like

Library/Models/QuotaFrameVersion.cs
joukevandermaas: No that's perfect, if they use it like that let's do it like that. But let's also make sure the capitalization is the same in the SDK code and not just in the json.

Library/Services/Implementation/NfieldQuotaService.cs
joukevandermaas: I think it's fine like this.

PR #207, opened 2020-07-20
new endpoint to save quota frame targets for etag

behnazbeaumont:


Library/Services/INfieldQuotaService.cs
```diff
@@ -25,10 +25,18 @@ namespace Nfield.Services
...
         /// <returns></returns>
         Task<IEnumerable<QuotaFrameVersion>> GetQuotaFrameVersionsAsync(string surveyId);
+
+        /// <summary>
+        /// Updates the survey quota targets for the specified quota frame version
+        /// </summary>
+        /// <param name="surveyId">The survey to set the quota targets for</param>
+        /// <param name="quotaETag">The quota frame version to set the targets for</param>
+        /// <param name="targets">The new quota frame targets</param>
+        Task SaveQuotaTargetsAsync(string surveyId, string quotaETag, IEnumerable<QuotaFrameLevelTarget> targets);
```
joukevandermaas, line 40: I guess I would prefer `Update` over `Save` because it seems more natural in C# code, but it doesn't really matter that much.

PR #209, opened 2020-07-21
Fixed name and type for etag

kefishev:


Library/Services/INfieldQuotaService.cs
```diff
@@ -43,8 +43,8 @@ public interface INfieldQuotaService
         /// Gets the specified version of the quota frame  
         /// </summary>
         /// <param name="surveyId">The survey id</param>
-        /// <param name="eTag">The version of the quota frame to retrieve</param>
+        /// <param name="Etag">The version of the quota frame to retrieve</param>
         /// <returns></returns>
-        Task<QuotaFrame> GetQuotaFrameAsync(string surveyId, long eTag);
+        Task<QuotaFrame> GetQuotaFrameAsync(string surveyId, string Etag);
```
joukevandermaas, line 48: I guess it should be `etag` right? then it's the same as the other one but using C# naming conventions.

WimDoze: When we want to be consistent it should stay eTag (see all other methods that have an ETag argument), but in QuotaFrameVersion it should be ETag

kefishev: I think Wim is right, in the api we also use eTag for arguments

PR #211, opened 2020-07-23
fixed route for updating quota frame targets

behnazbeaumont:


behnazbeaumont:
> This does not fix the bug; maybe it fixes another bug but the endpoint that doesn't work is still broken after this change.

this was one that @kefishev and I found, just to be sure is the one you came across INfieldQuotaService.GetQuotaFrameAsync?

Library/Services/Implementation/NfieldQuotaService.cs
```diff
@@ -45,7 +45,7 @@ public Task UpdateQuotaTargetsAsync(string surveyId, string eTag, IEnumerable<Qu
         {
             ValidateSurveyId(surveyId);
 
-            return Client.PutAsJsonAsync(EditingQuotaFrameTargetsUri(surveyId, eTag), targets)
+            return Client.PutAsJsonAsync(QuotaFrameTargetsUri(surveyId, eTag), targets)
                          .FlattenExceptions();
```
kefishev, line 49: It now points to the correct api controller. 
But there is another problem though. The type of etag is a string in SDK but in the controller it's a long

joukevandermaas:
@behnazbeaumont that's right.

Library/Services/Implementation/NfieldQuotaService.cs
behnazbeaumont: urgh so messy

behnazbeaumont:
> @behnazbeaumont that's right.

@joukevandermaas the test is failing probably because the eTag param to the api is a string and in the test we're passing a long

Library/Services/Implementation/NfieldQuotaService.cs
behnazbeaumont: @kefishev I talked with @joukevandermaas and it should be ok, it will be handled on the api side if etag is invalid.

PR #212, opened 2020-08-04
Versioned quota through public API

joukevandermaas:
Since it is now possible to change quota during fieldwork, some new endpoints were added to query previous frames like is possible in the Nfield Manager.

### DIFF
--- a/Library/Services/Implementation/NfieldQuotaService.cs
+++ b/Library/Services/Implementation/NfieldQuotaService.cs
@@ -1,0 +1,58 @@
+﻿using Newtonsoft.Json;
+using Nfield.Infrastructure;
+using Nfield.Models;
+using Nfield.Services;
+using Nfield.Extensions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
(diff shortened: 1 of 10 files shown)


Library/Services/Implementation/NfieldQuotaService.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// See <see cref="INfieldQuotaService.GetQuotaFrameVersionsAsync"/>
+        /// </summary>
+        public Task<IEnumerable<QuotaFrameVersion>> GetQuotaFrameVersionsAsync(string surveyId)
+        {
+            ValidateSurveyId(surveyId);
+
+            return Client.GetAsync(QuotaFrameVersionsUri(surveyId))
+                         .ContinueWith(task => task.Result.Content.ReadAsStringAsync().Result)
```
marcoNipo, line 36: I though we could do async await in the SDK already 🤔

joukevandermaas: We can, probably this was copied from an older service. I will see if we can refactor it.

PR #217, opened 2020-10-06
Add Email column property

pilarodriguez:
[AB#70786](https://niposoftware.visualstudio.com/Nfield/_workitems/edit/70786)

### DIFF
--- a/Library/Models/SurveyEmailSettings.cs
+++ b/Library/Models/SurveyEmailSettings.cs
@@ -41,5 +41,10 @@ namespace Nfield.Models
         /// </summary>
         public string Id { get; set; }
 
+        /// <summary>
+        /// The column used to store the emails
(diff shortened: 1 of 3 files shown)


joukevandermaas (commented):
Will there also be a change to `DomainEmailSettings`? If all surveys in the domain use the same email address column, that would be convenient to have for users.

pilarodriguez:
No, only at survey level. The parameter was already available in sample management, but we were missing the public api.

Actually, we wanted to have the EmailColumn as a fixed column, because now for joining the blacklist table, we first need to know which is the EmailColumn and the query is even more complicated than before. But PM decision was that we need to continue giving the customers the option to select their own EmailColumn per survey

joukevandermaas (approved):

PR #226, opened 2021-05-28
Disabled "warning as error" for debug

kefishev:


behnazbeaumont (commented):
@kefishev you need to set the TreatWarningsAsErrors to true for the release case

kefishev:
@behnazbeaumont If I do that, the release build will fail.

behnazbeaumont:
> > @behnazbeaumont If I do that, the release build will fail.
> 
> @kefishev yes I know, we should figure it out for all the ones with the false value for release
> @behnazbeaumont Do you mean to fix release errors?

if we want to make this change to all repos we have to fix the release, unless we decide to skip those that don't build.

behnazbeaumont:
> > > @behnazbeaumont If I do that, the release build will fail.
> > 
> > 
> > @kefishev yes I know, we should figure it out for all the ones with the false value for release
> > @behnazbeaumont Do you mean to fix release errors?
> 
> if we want to make this change to all repos we have to fix the release, unless we decide to skip those that don't build.
> @behnazbeaumont It will not be easy, the build generates 367 errors

@kefishev I just talked about it wit @joukevandermaas and he said the release build fix is out of scope and we should add our bit and leave the release as false

joukevandermaas (approved):

PR #234, opened 2021-07-19
Add get/set dial mode

WimDoze:
Add get/set dial mode api calls


Library/Services/Implementation/NfieldSurveysService.cs
```diff
@@ -581,4 +608,9 @@ internal class UpdateSurvey
         public string InterviewerInstruction { get; set; }
     }
 
+    internal class UpdateDialMode
```
behnazbeaumont, line 611: perhaps DialModeModel would be a better name since it's used in both set and get?

WimDoze: Changed the name

kefishev:
Don't forget to update version.txt. I would do it as part of this PR.

WimDoze:
> Don't forget to update version.txt. I would do it as part of this PR.

I always do that when I merge to master, need to check then anyway.

kefishev (approved):

PR #243, opened 2021-09-16
Localusers sdk

ozaragoza:
New endpoint and unit test


Library/Models/LocalUser.cs
```diff
@@ -77,4 +77,15 @@ public class LocalUser : LocalUserValues
...
         public DateTime? LastLogonDate { get; internal set; }
     }
+
+    /// <summary>
+    /// Model representing the settable values in a survey group.
+    /// </summary>
+    public class ChangePasswordLocalUser
+    {
+        /// <summary>
+        /// Password of the new user
```
AlbertoBelver, line 87: Is not a new user

Library/Services/Implementation/NfieldLocalUserService.cs
```diff
@@ -82,6 +82,17 @@ public async Task<LocalUser> UpdateAsync(string identityId, LocalUserValues mode
...
 
+        public async Task ResetAsync(string identityId, ChangePasswordLocalUser model)
+        {
+            if (model == null)
+                throw new ArgumentNullException(nameof(model));
+
+            var uri = new Uri(ConnectionClient.NfieldServerUri, $"LocalUsers/Password/{identityId}");
+            using (await ConnectionClient.Client.PatchAsJsonAsync(uri, model).ConfigureAwait(false))
+            {
+            }
```
AlbertoBelver, line 93: Why an empty using?

DfernandezNipo: I think for dispose it. It's done everywhere in the api

PR #245, opened 2021-09-17
Api: Add themes and templates endpoints

hcsauniere:
[AB#95768](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/95768)

PRs:
- https://github.com/NIPOSoftwareBV/nfield-source/pull/6832
- https://github.com/NIPOSoftware/Nfield-SDK/pull/245

Tests:
- [x] capi and templates
- [x] saas admin templates
- [x] Api integration tests


Library/Models/Template.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// <summary>
+    /// Holds all properties of a template
+    /// </summary>
+    public class Template
+    {
+
+        /// <summary>
+        /// Unique ID of the template
+        /// </summary>
+        [JsonProperty]
```
marcoNipo, line 30: why this attribute? Same for `Theme.Id`

marcoNipo (commented):
don't forget to update the version
https://github.com/NIPOSoftware/Nfield-SDK#versioning

Library/Models/Template.cs
marcoNipo: I see, but for which of [these cases](https://www.newtonsoft.com/json/help/html/serializationattributes.htm) do you need this attribute here?

```
- By default, the JSON property will have the same name as the .NET property. This attribute allows the name to be customized.
- JsonPropertyAttribute indicates that a property should be serialized when member serialization is set to opt-in.
- It includes non-public properties in serialization and deserialization.
- It can be used to customize type name, reference, null, and default value handling for the property value.
- It can be used to customize the NamingStrategy of the serialized property name.
- It can be used to customize the property's collection items JsonConverter, type name handling, and reference handling.
```

Tests/Services/NfieldTemplatesServiceTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+            };
+            var mockedNfieldConnection = new Mock<INfieldConnectionClient>();
+            var mockedHttpClient = new Mock<INfieldHttpClient>();
+            mockedHttpClient
+                .Setup(client => client.GetAsync(new Uri(ServiceAddress, "Templates/")))
+                .Returns(
+                    Task.Factory.StartNew(
+                        () =>
+                            new HttpResponseMessage(httpStatusCode)
```
marcoNipo, line 75: a bit weird that `httpStatusCode` is `HttpStatusCode.NotFound` with some content, and that `QueryAsync` returns valid objects, but I see the SDK pretty much always ignores the the response code.
Maybe use OK instead for the test

marcoNipo (approved):

PR #251, opened 2021-10-07
Yellow/ci public api user logs

DfernandezNipo:


Library/Models/LogQueryModel.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Class to query logs
+    /// </summary>
+    public class LogQueryModel
+    {
+        public DateTime From { get; set; }
+        public DateTime To { get; set; }
```
joukevandermaas, line 26: Same comment here, I would prefer `FromUtc` and `ToUtc` for the sake of clarity.

PR #253, opened 2021-10-07
Add NfieldSurveyLanguageTranslationsService and tests

MartinRijks:


Library/Services/INfieldSurveyLanguageTranslationsService.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    public interface INfieldSurveyLanguageTranslationsService
+    {
+        #region CRUD on Language Translations
+
+        /// <summary>
+        /// Gets all languages and their translations for a survey, as a queryable object.
+        /// </summary>
+        /// <param name="surveyId">The survey for which language names and ids are returned</param>
+        /// <exception cref="T:System.AggregateException"></exception>
```
hcsauniere, line 34: We should probably remove all the empty comments in the file about the exceptions (or fill them)

Library/Services/INfieldSurveyLanguageTranslationsService.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// The aggregate exception can contain:
+        /// <exception cref="Nfield.Exceptions.NfieldErrorException"></exception>
+        /// <exception cref="Nfield.Exceptions.NfieldHttpResponseException"></exception>
+        Task<IQueryable<SurveyLanguageTranslations>> QueryAsync(string surveyId);
+
+        /// <summary>
+        /// Adds a new language and all of its translations.
+        /// </summary>
+        /// <param name="surveyId">the survey to add this language to</param>
+        /// <param name="languageId">A new id for the language to add</param>
```
hcsauniere, line 44: The add language shouldn't have a languageId

Library/Services/Implementation/NfieldSurveyLanguageTranslationsService.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (languageId == null)
+                throw new ArgumentNullException("languageId");
+            if (languageId.Trim().Length == 0)
+                throw new ArgumentException("LanguageId cannot be empty");
+        }
+
+        private Uri LanguageTranslationsApi(string surveyId, string languageId = null)
+        {
+            var path = new StringBuilder();
+            path.AppendFormat("Surveys/{0}/Languages/Translations", surveyId);
```
hcsauniere, line 119:
```suggestion
            path.AppendFormat("Surveys/{0}/LanguageTranslations", surveyId);
```

Library/Services/INfieldSurveyLanguageTranslationsService.cs
MartinRijks: It's part of the model?

I thought we agreed in our API plans that IDs would be taken out of the models as they are usually used as parameter in the URL REST path

Library/Services/INfieldSurveyLanguageTranslationsService.cs
MartinRijks: Most of the interfaces have these listed, I stuck with it.

Tests/Services/NfieldSurveyLanguageTranslationTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            var content = new StringContent(JsonConvert.SerializeObject(translations));
+            mockedHttpClient
+                .Setup(client => client.PatchAsJsonAsync(new Uri(ServiceAddress, $"Surveys/{SurveyId}/LanguageTranslations/{languageId}"), translations))
+                .Returns(CreateTask(HttpStatusCode.OK, content));
+
+            var target = new NfieldSurveyLanguageTranslationsService();
+            target.InitializeNfieldConnection(mockedNfieldConnection.Object);
+
+            var actual = target.UpdateAsync(SurveyId, languageId, translations).Result;
+            target.UpdateAsync(SurveyId, languageId, translations);
```
hcsauniere, line 162:
```suggestion
```

Tests/Services/NfieldSurveyLanguageTranslationTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            Assert.Throws<ArgumentException>(() => UnwrapAggregateException(target.AddAsync(string.Empty, new SurveyLanguageTranslations())));
+        }
+
+        [Fact]
+        public void TestAddAsync_ServerAcceptsLanguageTranslations_ReturnsLanguageTranslations()
+        {
+            var translations = new SurveyLanguageTranslations()
+            {
+                ButtonNext = "Next",
+                ButtonBack = "Back",
```
AlbertoBelver, line 104: I don't know if it would be necessary add all the properties to test if the SerializeObject & DeserializeObject works for all

MartinRijks: There are many more properties, I think testing one or two is enough - we should not have to test whether (de)serialization works as it should. Or are you expecting issues with it?

PR #255, opened 2021-10-07
domain response codes methods

DfernandezNipo:


Library/Services/Implementation/NfieldDomainResponseCodesService.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            get { return ConnectionClient.Client; }
+        }
+
+
+
+        /// <summary>
+        /// <see cref="INfieldDomainResponseCodesService.QueryAsync(string)"/>
+        /// </summary>
+        public Task<IQueryable<DomainResponseCode>> QueryAsync(string domainId)
```
hcsauniere, line 45: The domain should not be part of the queries

hcsauniere (approved):

PR #275, opened 2021-11-09
Add Windows Forms sample using AAD authentication

bjorncoltof:
This PR adds a sample on how to use the SDK in  a Windows Forms application while using AAD authentication.

The existing console sample was moved into its own folder and renamed to 'ConsoleApplication'

Tasks

- [X] Add Windows Forms sample using AAD
- [x] Document process around app creation in the Azure portal


Examples/WindowsFormsApplication/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,37 @@
...
+        private static ApplicationConfiguration InitializeConfiguration()
+        {
+            var root = new ConfigurationBuilder()
+                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
+                .AddUserSecrets<Program>()
+                .Build();
+
+            return new ApplicationConfiguration
+            {
+                ClientId = root["Nfield:ClientId"],
```
LNieuwmegen, line 29: The "client" itself is something "external" to nfield I would say, so putting it in the configuration file under the Nfield namespace might be confusing. Maybe have a Nfield level where we put all nfield relevant config and a separate level where we put the "3rd party app" stuff. This might be over the top :-P.

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            client.UserTokenCache.SetBeforeAccess(BeforeAccessNotification);
+            client.UserTokenCache.SetAfterAccess(AfterAccessNotification);
+
+            return client;
+        }
+
+        private async Task<string> ProvideTokenAsync()
+        {
+            var result = await AuthenticateAsync((status) => { });
```
LNieuwmegen, line 45: Maybe some extra context here on why an empty anonymous function is sufficient.

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+            DomainName = domainName;
+
+            NfieldConnection = NfieldConnectionFactory.Create(ApplicationConfiguration.Current.NfieldApiUri);
+            NfieldConnection.RegisterTokenProvider(domainName, ProvideTokenAsync);
+        }
+
+        private IPublicClientApplication CreateClientApp()
+        {
+            var client = PublicClientApplicationBuilder.Create(ApplicationConfiguration.Current.ClientId)
+                .WithRedirectUri("http://localhost")
```
LNieuwmegen, line 33: This is a native client right? do they need redirecturi's to localhost? I thought that works differently (with some weird strings and stuff)

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+                await ClientApp.RemoveAsync(accounts.First());
+                accounts = (await ClientApp.GetAccountsAsync()).ToList();
+            }
+        }
+
+        private byte[] RestoreTokenCacheFromRegistry()
+        {
+            using (RegistryKey key = Registry.CurrentUser.CreateSubKey(@"SOFTWARE\NIPO Samples"))
+            {
+                var value = key.GetValue("TokenCache");
```
LNieuwmegen, line 65: Is it common practice to store token cache in  registry? shouldnt we use the special windows sdk for save consolidated app storage (since tokens are critical/security sensitive). I think there are some secure data api's or something in windows to deal better with this.

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+            var accounts = await ClientApp.GetAccountsAsync();
+            var account = accounts.FirstOrDefault();
+
+            try
+            {
+                return await ClientApp.AcquireTokenSilent(Scopes, account).ExecuteAsync();
+            }
+            catch (MsalUiRequiredException)
+            {
+                return null;
```
LNieuwmegen, line 105: Not sure how far we want to take this sample but maybe we should add some logging as a best practice :-D

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        public async Task<AuthenticationResult> AuthenticateAsync(Action<string> statusCallback)
+        {
+            var accounts = await ClientApp.GetAccountsAsync();
+            var account = accounts.FirstOrDefault();
```
LNieuwmegen, line 112: Is it encouraged to always take the first account? why? in logout i saw also an enumeration of multipe accounts. I am not super familiar with this new libs yet so I wonder if this is good practice :-).

Examples/WindowsFormsApplication/Nfield.cs
```diff
@@ -0,0 +1,138 @@
...
+            try
+            {
+                return await ClientApp.AcquireTokenSilent(Scopes, account).ExecuteAsync();
+            }
+            catch (MsalUiRequiredException mure)
+            {
+                statusCallback(mure.Message);
+            }
+
+            var authResult = await ClientApp.AcquireTokenInteractive(Scopes)
```
LNieuwmegen, line 123: This kind of stuff definately can use some documentation (why are we doing interactive login at this point). If you get MsalUiRequiredException it usually means that e.g. consent should be given but that might not be fully clear to the user here without some context.

Examples/WindowsFormsApplication/readme.md
```diff
@@ -0,0 +1,127 @@
...
+![Add a platform](./images/AddPlatform.png)
+
+### _Add support for `Mobile and desktop applications`_
+![Mobile and desktop applications](./images/AddMobileDesktopApplication.png)
+
+### _Setup `Redirect URIs`_
+
+In order to support the ability to show the browser to select the account to use for logging in to Nfield the following URIs should be configured. The `http://localhost` URI is needed when using .Net Core, it will allow the application to capture the result of the login from the browser.
+
+![Redirect URIs](./images/RedirectUris.png)
```
LNieuwmegen, line 39: do we also need the /common/nativeclient thingy?

Examples/WindowsFormsApplication/readme.md
```diff
@@ -0,0 +1,127 @@
...
+
+### _Select the Nfield Public API_
+
+Look up the Nfield Public API in the `APIs my organizations uses` and store the its Application ID.
+
+![Select Nfield Public API](./images/SelectNfieldPublicAPI.png)
+
+### _Add all Nfield permissions_
+
+![Add all permissions](./images/AddAllNfieldPublicAPIPermissions.png)
```
LNieuwmegen, line 60: I would not recommend to add "all permissions" from a "zero trust" "least privilage" standpoint

Examples/WindowsFormsApplication/readme.md
```diff
@@ -0,0 +1,127 @@
+# Windows Forms sample
```
LNieuwmegen, line 1: A lot of the questions from above are answered here I think, maybe good to at critical points add some links from the code to specific sections in this doc.

Examples/WindowsFormsApplication/readme.md
```diff
@@ -0,0 +1,127 @@
...
+private static string CreateScope(string scope) => $"{ApplicationConfiguration.Current.NfieldApiApplicationId}/{scope}";
+```
+Nfield only supports scopes based on the Nield Public API application ID.
+The name as presented in the permissions UI in the Azure portal cannot not be used.
+
+Now after launching the application and logging in for the first time the user is asked to consent giving access to the application. For the sample application the screen will look like this:
+
+![Consent](./images/Consent.png)
+
+After giving consent the user should be presented with a screen that shows the surveys in the domain.
```
LNieuwmegen, line 127: It would be nice to also link to some formal MS docs if people would like to know more (especially on the permission part/app config).

LNieuwmegen (commented):
Nice dude, some notes below.

Examples/WindowsFormsApplication/Nfield.cs
LNieuwmegen: What is also funny here is that you enabled the "native client platform" on the app registration yet in this appbuilder nothing points to that it is a native client (that cannot  keep secrets). Not super familiar with this new apis but it made me scratch my head a bit :-D

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: It's what they use in all the sample. This library also supports multiple account, but this app only ever uses one, however the API is geared towards multiple.

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: It is more likely that you simply need to authenticate, I'll add comments.

Examples/WindowsFormsApplication/readme.md
bjorncoltof: If you use regular .Net you can use that one, for .Net Core (and up) this is needed as there they don't support the embedded browser flow.

Examples/WindowsFormsApplication/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+
+namespace WindowsFormsApplication
+{
+    public class ApplicationConfiguration
+    {
+        public string ClientId { get; private set; }
+        public string Tenant { get; private set; }
+        public string NfieldApiApplicationId { get; private set; }
+        public Uri NfieldApiUri { get; internal set; }
```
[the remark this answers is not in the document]
bjorncoltof: Now actually they don't as the beta and production app use the same app, but a different URL.

Examples/WindowsFormsApplication/ApplicationConfiguration.cs
bjorncoltof: I'll put ClientId and TenantId in an `Application` section

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: I'll add a comment, I don't intend to make this a best practise app all round, just the required pieces to get AAD authentication and authorization up and running.

Examples/WindowsFormsApplication/readme.md
bjorncoltof: Upadting the text

Examples/WindowsFormsApplication/ApplicationConfiguration.cs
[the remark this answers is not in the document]
LNieuwmegen: Ohhh wait, i thought this was the AAD AppId and AAD AppUri, but it is actually the real url where the services are hosted :-D Maybe call that ApiEndpoint or something :-D

Examples/WindowsFormsApplication/readme.md
bjorncoltof: I have sprinkled some nice links throughout the document.

Examples/WindowsFormsApplication/ApplicationConfiguration.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Microsoft.Extensions.Configuration;
+using System;
+
+namespace WindowsFormsApplication
+{
+    public class ApplicationConfiguration
+    {
+        public string ClientId { get; private set; }
+        public string Tenant { get; private set; }
```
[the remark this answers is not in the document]
bjorncoltof: Renamed to TenantId

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: Added comment in code and extracted the noop to a local variable

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: The readme explains why this URI is used, including links to ms docs.

Examples/WindowsFormsApplication/Nfield.cs
bjorncoltof: Registry is quite common for client apps, I now added data protection scoped to the current user to it, so only that user can decrypt the token data.

Examples/WindowsFormsApplication/readme.md
bjorncoltof: I added ms docs link in various places

LNieuwmegen (approved):

PR #279, opened 2021-11-11
#98467 Public API: Add endpoints for Domain Languages

MartinRijks:
Story: [#98467](https://dev.azure.com/niposoftware/Nfield/_sprints/taskboard/Team%20Yellow/Nfield/Release%202021/2021%20Q4%20Sprint%2003?workitem=98467)

Related PRs:
nfield-source: https://github.com/NIPOSoftwareBV/nfield-source/pull/6978
project-glu-api: https://github.com/NIPOSoftwareBV/project-glu-api/pull/3572

Tests run:
- [x] `Nfield.SDK.Tests` (all)


Library/Services/Implementation/NfieldDomainLanguageTranslationsService.cs
```diff
@@ -0,0 +1,102 @@
...
+            ConnectionClient = connection;
+        }
+
+        #endregion
+
+        #region Implementation of INfieldLanguageTranslationsService
+
+        public Task<IQueryable<Language>> QueryAsync()
+        {
+            return Client.GetAsync(LanguageTranslationsApi())
```
LNieuwmegen, line 56: Funny, do we as of today still have to deal with task continuation instead of async/await because of clients that depend on a sufficiently old version of .NET?

LNieuwmegen (approved):

Library/Services/Implementation/NfieldDomainLanguageTranslationsService.cs
MartinRijks: To be honest, I don't know. :-/  We took implementation of other services as an example and ran along with that (so, it's just stupid copypasta). It might be nice to throw a little investigation about .NET versions in use by our API consumers.

PR #284, opened 2021-11-26
Yellow/response codes counts

DfernandezNipo:


hcsauniere (commented):
Same comments in the naming & the PR as in the nfield-source repo. I will review it once those changes are done

PR #292, opened 2021-12-02
Public API: New endpoint: Get Quota Frame

ozaragoza:
Access from the survey service to the new Endpoint

[AB#100643](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/100643)

### DIFF
--- a/Postman/SurveyQuotas.postman_collection.json
+++ b/Postman/SurveyQuotas.postman_collection.json
@@ -60,7 +60,7 @@
 			"response": []
 		},
 		{
-			"name": "Get QuotaTargets Etag",
+			"name": "Get QuotaTargets by Etag",
 			"event": [
 				{
 					"listen": "test",
--- a/Postman/SurveyQuotas.postman_collection.json
+++ b/Postman/SurveyQuotas.postman_collection.json
@@ -1,0 +1,118 @@
+{
+	"info": {
+		"_postman_id": "b5aa0cd7-e932-4317-bbae-9c7ccf972cff",
+		"name": "SurveyQuotas",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
+		{
+			"name": "Get QuotaTargets",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"Status code is 200\", function () {",
+							"    pm.response.to.have.status(200);",
+							"});",
+							"",
+							"pm.test(\"Status code name has string\", function () {",
+							"    pm.response.to.have.status(\"OK\");",
+							"});",
+							"",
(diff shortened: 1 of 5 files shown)


Postman/SurveyQuotas.postman_collection.json
```diff
@@ -0,0 +1,118 @@
+{
+	"info": {
+		"_postman_id": "b5aa0cd7-e932-4317-bbae-9c7ccf972cff",
+		"name": "SurveyQuotas",
```
hcsauniere, line 4: The file name should be `SurveyQuota `or `SurveyQuotaTargets` or part of the `Surveys` collection (as the code has been added to the that service)

Postman/SurveyQuotas.postman_collection.json
```diff
@@ -0,0 +1,118 @@
...
+						"{{SurveyId}}",
+						"QuotaTargets"
+					]
+				},
+				"description": "This method retrieves the QuotaTargets from a Survey."
+			},
+			"response": []
+		},
+		{
+			"name": "Get QuotaTargets Etag",
```
hcsauniere, line 63:
```suggestion
			"name": "Get QuotaTargets by Etag",
```

Tests/Services/NfieldSurveysServiceTests.cs
```diff
@@ -223,28 +225,169 @@ public void TestUploadInterviewerInstructionsAsync_ValidByteArray_FileUpload()
...
+            var actualQuotaLevel = target.QuotaQueryAsync(SurveyId).Result;
             mockedHttpClient.Verify(hc => hc.GetAsync(It.IsAny<Uri>()), Times.Once());
             Assert.Equal(expectedQuotaLevel.Id, actualQuotaLevel.Id);
             Assert.Equal(expectedQuotaLevel.Name, actualQuotaLevel.Name);
+        }
+
+        [Fact]
+        public void TestQuotaTargetsQueryAsync_ServerReturnsQuery_ReturnsListWithQuotaFrame()
+        {
+            const string QuotaFrameId = "quotaId";
```
hcsauniere, line 251: Most (all?) of these constants are only used once in the test (same for the next one). From my point of view, adding them as independent variables in this case doesn't help and they could be removed

hcsauniere (commented):
My feeling is that we are over-extending the SurveysService, but I'm fine leaving it as is for now.

hcsauniere (approved):
My ideas on the tests was more to hardcode the value than to create a constant for it (at the end it's harder to maintain with so many constants), but anyway I'm fine as it is

ozaragoza:
In fact, I only did 1 test... The rest of the tests were already using the consts like that... I have done a refactor to reuse them as much as possible.

PR #297, opened 2021-12-10
fix postman

DfernandezNipo:


Postman/LocalUsers.postman_collection.json
```diff
@@ -1 +1,145 @@
...
\ No newline at end of file
+{
+  "info": {
+    "_postman_id": "e788e1cb-c96d-4983-a2e0-d4c20f9996d1",
+    "name": "Local users",
```
MartinRijks, line 4: I think this is the name as it appears in the collection, but not sure (that's why I asked you to check :-)). If so, all the other names use capitalization and no spaces and the name should be `LocalUsers`.

Postman/templates_and_themes.postman_collection.json
```diff
@@ -1 +1,90 @@
...
\ No newline at end of file
+{
+  "info": {
+    "_postman_id": "374e6bb5-a2f8-4c52-a708-abcfec06e3bb",
+    "name": "Themes & Templates",
```
MartinRijks, line 4: Same as previous, but then just `Templates` because that is the controller.

MartinRijks (approved):

Postman/LocalUsers.postman_collection.json
DfernandezNipo: It's the folder name. I will change them.

PR #315, opened 2022-02-23
Yellow/fix postman collection

DfernandezNipo:
[AB#105180](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/105180)

### DIFF
--- a/Postman/Nfield Public API.postman_collection.json
+++ b/Postman/Nfield Public API.postman_collection.json
@@ -851,7 +851,7 @@
 													}
 												},
 												"url": {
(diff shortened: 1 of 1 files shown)


hcsauniere (changes requested):
Very nice to standardize it as initially proposed. 
Just a couple of errors that I have seen:
- Sampling Point Download Image: Path Variables not set
- Languages & LanguagesTranslations endpoints: paths not properly updated where the LanguageId is still part of it

PR #318, opened 2022-03-04
#AB104569: Api.Domain: Remove Classic Code (from Domain endpoints)

MartinRijks:
[#AB 104569](https://dev.azure.com/niposoftware/Nfield/_sprints/taskboard/Team%20Yellow/Nfield/Release%202022/2022%20P1S3?workitem=104569)

To remove some dependencies on classic code we need to introduce a slightly updated service (changing the existing service would be a breaking change). The `CapiInterviewersService` replaces the `InterviewersService`, which is now obsolete.

See also changes in:
[Public API](https://github.com/NIPOSoftwareBV/nfield-source/pull/7232)

Tests run:
- [x] Unit tests only, package will be integration tested in the public API.


Library/Models/InterviewerChanged.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using Newtonsoft.Json;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Holds key properties of the created interviewer
+    /// </summary>
+    public class InterviewerChanged
```
AlbertoBelver, line 24: The model **InterviewerChanged** is used to create, update and password update, so this name is not the best one.
I suggest:
- InterviewerResponse
- InterviewerIdentifier
- InterviewerKeys

Also you should change the summary to add ` create update or password update`

Library/Services/Implementation/NfieldCapiInterviewersService.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            var updatedInterviewer = new UpdateCapiInterviewer
+            {
+                EmailAddress = interviewer.EmailAddress,
+                FirstName = interviewer.FirstName,
+                LastName = interviewer.LastName,
+                TelephoneNumber = interviewer.TelephoneNumber,
+                IsSupervisor = interviewer.IsSupervisor
+            };
+
```
AlbertoBelver, line 81: You can use an anonymous class and remove the internal class UpdateCapiInterviewer
```suggestion
            var updatedInterviewer = new
            {
                EmailAddress = interviewer.EmailAddress,
                FirstName = interviewer.FirstName,
                LastName = interviewer.LastName,
                TelephoneNumber = interviewer.TelephoneNumber,
                IsSupervisor = interviewer.IsSupervisor
            };

```

Library/Services/Implementation/NfieldInterviewersService.cs
```diff
@@ -29,13 +29,16 @@ namespace Nfield.Services.Implementation
...
+
     internal class NfieldInterviewersService : INfieldInterviewersService, INfieldConnectionClientObject
     {
         #region Implementation of INfieldInterviewersService
 
+        const string serviceObsoleteMessage = "NfieldInterviewersService is obsolete, please use NfieldCapiInterviewersService instead.";
         /// <summary>
         /// See <see cref="INfieldInterviewersService.AddAsync"/>
         /// </summary>
+        [Obsolete(serviceObsoleteMessage)]
```
AlbertoBelver, line 41: Put it at class level should be enough
```suggestion
```

Library/Services/Implementation/NfieldCapiInterviewersService.cs
ozaragoza: But it's better for the testing, so you expect a certain class in the mock class

marcoNipo (approved):

PR #322, opened 2022-03-09
Password back in the editable model, remove password specific methods, cleanup

MartinRijks:
Fixes that you cannot set a password when creating / updating the interviewer. Removes separately updating the password (you can use Update for that). When retrieving an interviewer, the password is not provided (so you can only set it, not read it).

[#105899](https://dev.azure.com/niposoftware/Nfield/_sprints/taskboard/Team%20Yellow/Nfield/Release%202022/2022%20P1S4?workitem=105899)

### DIFF
--- a/Library/Models/CapiInterviewers/CapiInterviewer.cs
+++ b/Library/Models/CapiInterviewers/CapiInterviewer.cs
@@ -15,7 +15,7 @@
 
 namespace Nfield.Models
 {
-    public class CapiInterviewer : EditCapiInterviewer
+    public class CapiInterviewer
     {
         /// <summary>
         /// Unique Id of Interviewer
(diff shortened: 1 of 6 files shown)


hcsauniere (commented):
I don't know the API implementation at this point. If it's the one in the PR: https://github.com/NIPOSoftwareBV/nfield-source/pull/7237 then the sdk and the api are not aligned as the userName seems that can only be set in the creation. I'm fine merging everything into the CIs and I review both PRs (nfield-source & nfield-sdk) entirely as the most important thing is that they are aligned.

MartinRijks:
> I don't know the API implementation at this point. If it's the one in the PR: [NIPOSoftwareBV/nfield-source#7237](https://github.com/NIPOSoftwareBV/nfield-source/pull/7237) then the sdk and the api are not aligned as the userName seems that can only be set in the creation. I'm fine merging everything into the CIs and I review both PRs (nfield-source & nfield-sdk) entirely as the most important thing is that they are aligned.

The original API implementation (Interviewers) supported changing the username:

https://yellow-api.niposoftware-dev.com/help/api/patch-v1-interviewers-interviewerid

I guess the changes provided by Oscar changed that so that it is no longer allowed. Question is, what do we want? I guess allowing to change the username is tricky as it effectively changes the sign-in username of the interviewer.

In the same vain, do you agree that we can remove the password-only change in the API? I noticed now that we use PUT to change the password and PATCH to change the interviewer. I think it's a weird difference (we can just use PATCH to set the password).

hcsauniere:
We should focus into bringing the same functionality that glu-api brings to the public-api, just that.
We can take other considerations in the future, but not as part of this

PR #323, opened 2022-03-11
CapiInterviewer service uses two endpoints, not one. Fix

MartinRijks:


hcsauniere (commented):
I don't get it. If you do this changes, the code would be targeting the same endpoints as the existing InterviewersService. In that case it doesn't make sense to add the code here (unless we mark the other one as obsolete, but then we would need a reason).
On Tuesday all of this must be clear or we will have again a problem merging the story.

MartinRijks:
@hcsauniere The problem is that the interviewersservice in the SDK makes use of two endpoints, one of which we changed into CapiInterviewers. The other endpoint is Interviewers/{interviewerId}/Offices which has a separate controller in the API (InterviewerOffices).

The only alternative I see is that we add a new route pointing to that controller, but technically it is the same as using the old route. And I don't know if you can have two routes pointing to the same controller, because we need to keep the old one for backward compatibility.

Library/Services/Implementation/NfieldCapiInterviewersService.cs
```diff
@@ -210,6 +156,12 @@ private Uri CapiInterviewersApi
         }
         #endregion
 
+        // Required for interviewer offices operations
+        private Uri InterviewersApi
```
hcsauniere, line 160: This probably is not needed

hcsauniere (commented):
One comment about a method that is probably not needed but you can consider it approved

hcsauniere (approved):

PR #334, opened 2022-06-01
Yellow/fix delete interview

DfernandezNipo:
Fix calling to delete interview data.


hcsauniere (commented):
As we have commented, I think that it would be better to update the Api to return the DeletedTotal instead

PR #337, opened 2022-06-10
Spare sp sdk

DfernandezNipo:
[AB#110891](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/110891)


Library/Services/INfieldSamplingPointsService.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Nfield.Models;
+
+namespace Nfield.Services
+{
+    /// <summary>
+    /// Represents a set of methods to manage sampling points.
+    /// </summary>
+    public interface INfieldSamplingPointsService
```
hcsauniere, line 26: As we talked before, the surveysService already contains sampling point endpoints. Could be good to put them as obsolete (or something like that)

PR #342, opened 2022-08-03
Add new endpoints to sdk

hcsauniere:
[AB#114077](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/114077)

### DIFF
--- a/Library/Services/Implementation/NfieldSurveyDataService.cs
+++ b/Library/Services/Implementation/NfieldSurveyDataService.cs
@@ -23,7 +23,7 @@ namespace Nfield.Services.Implementation
                 throw new ArgumentNullException(nameof(surveyDownloadDataRequest));
             }
 
-            var uri = new Uri(ConnectionClient.NfieldServerUri, $"surveys/{surveyDownloadDataRequest.SurveyId}/data");
+            var uri = new Uri(ConnectionClient.NfieldServerUri, $"Surveys/{surveyDownloadDataRequest.SurveyId}/data");
 
             return Client.PostAsJsonAsync(uri, surveyDownloadDataRequest)
                          .ContinueWith(task => task.Result.Content.ReadAsStringAsync().Result)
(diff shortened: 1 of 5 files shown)


Library/Services/Implementation/NfieldSurveyDataService.cs
```diff
@@ -40,14 +74,17 @@ public void InitializeNfieldConnection(INfieldConnectionClient connection)
...
         #endregion
 
-        private INfieldHttpClient Client
+        private static void CheckSurveyId(string surveyId)
         {
-            get { return ConnectionClient.Client; }
+            if (surveyId == null)
+                throw new ArgumentNullException(nameof(surveyId));
+            if (surveyId.Trim().Length == 0)
+                throw new ArgumentException("surveyId cannot be empty");
```
AlbertoBelver, line 82: if (!Guid.TryParse(surveyId, out var newGuid))
        throw new ArgumentException("surveyId invalid format");

Library/Services/Implementation/NfieldSurveyDataService.cs
```diff
@@ -40,14 +74,17 @@ public void InitializeNfieldConnection(INfieldConnectionClient connection)
...
         }
 
-        private Uri SurveyDataUrl(string surveyId)
+        private INfieldHttpClient Client
```
AlbertoBelver, line 85:
```suggestion
        private INfieldHttpClient Client => ConnectionClient.Client;
```

AlbertoBelver (approved):

Library/Services/Implementation/NfieldSurveyDataService.cs
hcsauniere: we don't check for Ids being a guid anywhere. If we go in that direction it should be something more generic. The surveyId length zero is a check to generate a valid url mainly

PR #351, opened 2022-10-18
Add request to SDK

basbakkernipo:
[AB#117621](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/117621)

### DIFF
--- a/Library/Models/Request.cs
+++ b/Library/Models/Request.cs
@@ -1,0 +1,66 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System.Collections.Generic;
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Configuration for a request
+    /// </summary>
+    public class Request
+    {
+        /// <summary>
+        /// The id of the request
+        /// </summary>
+        public int Id { get; set; }
+
+        /// <summary>
+        /// Name of the request used in the *REQUEST command
+        /// </summary>
+        /// <returns></returns>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Endpoint (URI) of the request
+        /// </summary>
+        public string Uri { get; set; }
+
+        /// <summary>
+        /// The template of the payload
+        /// </summary>
+        public string PayloadTemplate { get; set; }
+
+        /// <summary>
+        /// The method of the request
+        /// </summary>
+        public RequestHttpMethod RequestHttpMethod { get; set; }
+
+        /// <summary>
+        /// Optional description of the request
+        /// </summary>
+        public string Description { get; set; }
+
+        /// <summary>
+        /// Optional help URI towards extensive documentation for the request
+        /// </summary>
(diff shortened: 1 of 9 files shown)


Library/Models/Request.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        /// <summary>
+        /// The id of the request
+        /// </summary>
+        public int Id { get; set; }
+
+        /// <summary>
+        /// Name of the request used in the *REQUEST command
+        /// </summary>
+        /// <returns></returns>
```
neilboyd, line 33:
```suggestion
```

Library/Services/INfieldRequestsService.cs
```diff
@@ -0,0 +1,51 @@
...
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using Nfield.Models;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Nfield.Services
+{
+    public interface INfieldRequestsService
```
neilboyd, line 22: comment on interface

Library/Services/INfieldRequestsService.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Removes the request.
+        /// </summary>
+        Task RemoveAsync(Request request);
+
+        /// <summary>
+        /// Updates the request.
+        /// 
+        /// All properties are updated.
+        /// Existing headers not present in 'externalApi' are deleted (matched on ExternalApiHeader.HeaderId)
```
neilboyd, line 44: copy-paste? `externalApi`

Library/Services/Implementation/NfieldRequestsService.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            return Client
+                .PostAsJsonAsync(RequestsApi, request)
+                .ContinueWith(task => task.Result.Content.ReadAsStringAsync().GetAwaiter().GetResult())
+                .ContinueWith(task => JsonConvert.DeserializeObject<Request>(task.GetAwaiter().GetResult()))
```
neilboyd, line 50: it's doing it sync? - `GetResult`

Postman/Nfield Public API.postman_collection.json
```diff
@@ -1,11 +1,11 @@
...
-		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
-		"_exporter_id": "631405"
-	},
+  "info": {
+    "_postman_id": "14c92e98-aba2-47c0-be41-5626e5d62b32",
```
neilboyd, line 3: does it matter that you've changed the id?

Library/Services/INfieldRequestsService.cs
basbakkernipo: Yep, I would swear I changed that...

Library/Services/Implementation/NfieldRequestsService.cs
basbakkernipo: Yep. Copy paste. Looks like the whole service works like this.

Postman/Nfield Public API.postman_collection.json
basbakkernipo: Don't think so. I think it's just an instance id of the postman client.

Library/Services/Implementation/NfieldRequestsService.cs
neilboyd: 😢

neilboyd (approved):

Library/Services/Implementation/NfieldRequestsService.cs
WimDoze: The whole service works like this because people keep copying it.
When you look at the NfieldCatiInterviewersService you can see that it is not necessary, but I think it is the only one.

basbakkernipo:
> Do not like not using await, but ok, if you want to.

I don't want to, it's how this package works. =(

WimDoze:
> > Do not like not using await, but ok, if you want to.
> 
> I don't want to, it's how this package works. =(

It is how the old stuff worked and people keep copying it, you can easily use await as the newer services do, but ok

PR #359, opened 2022-10-20
Support download blacklist

mvdr:
what about the _postman_id in the postman collection json file? How does that work?


hnuguse (approved):
I guess that postman id is also some identifier for the collection version. But I'm not sure if we ever use that.

PR #352, opened 2022-10-18
[AB#116867] Story: Implement configuration for *REQUEST in public API

basbakkernipo:
[AB#116867](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/116867)

Related PR:
- https://github.com/NIPOSoftwareBV/project-glu-api/pull/4391
- https://github.com/NIPOSoftware/Nfield-SDK/pull/352
- https://github.com/NIPOSoftwareBV/nfield-source/pull/7800

PR #361, opened 2022-10-27
[AB#116814] API for Downloading  blacklist email addresses

mvdr:
See also [nfield-source](https://github.com/NIPOSoftwareBV/nfield-source/pull/7850)

[AB#116814](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/116814)

### DIFF
--- a/Library/Services/INfieldBlacklistService.cs
+++ b/Library/Services/INfieldBlacklistService.cs
@@ -36,7 +36,6 @@ namespace Nfield.Services
         /// <summary>
         /// Downloads the blacklist as a tab delimited string, that can be stored as csv file.
         /// </summary>
-        /// <returns></returns>
         Task<string> GetAsync();
     }
 }
--- a/Library/Services/INfieldBlacklistService.cs
+++ b/Library/Services/INfieldBlacklistService.cs
@@ -32,5 +32,11 @@ namespace Nfield.Services
         /// <exception cref="ArgumentException"></exception>
         /// <returns><see cref="BlacklistUploadStatus"/></returns>
         Task<BlacklistUploadStatus> PostAsync(string blacklist);
+
+        /// <summary>
+        /// Downloads the blacklist as a tab delimited string, that can be stored as csv file.
+        /// </summary>
+        /// <returns></returns>
+        Task<string> GetAsync();
     }
 }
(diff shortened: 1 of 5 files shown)


Library/Services/Implementation/NfieldBlacklistService.cs
```diff
@@ -29,6 +29,15 @@ namespace Nfield.Services.Implementation
     /// </summary>
     internal class NfieldBlacklistService : INfieldBlacklistService, INfieldConnectionClientObject
     {
+        public Task<string> GetAsync()
+        {
+            var uri = BlacklistUrl();
+
+            return Client.GetAsync(uri)
+                .ContinueWith(task => task.Result.Content.ReadAsStringAsync().Result)
```
gkappen, line 37: Isn't this potentially dangerous (using `Result` on an async method). This should be awaited I think

gkappen (commented):
I have doubts

Library/Services/Implementation/NfieldBlacklistService.cs
gkappen: Why is this method not simply `async`?
``` c#
var result = await Client.GetAsync(uri);
var blacklist = await result.Content.ReadAsStringAsync();

return blacklist;
```
Or am I missing some vital bit of info?

neilboyd: We also copy-pasted some old code in our PR #352, but then changed it to use `await`.

Library/Services/INfieldBlacklistService.cs
```diff
@@ -32,5 +32,11 @@ public interface INfieldBlacklistService
         /// <exception cref="ArgumentException"></exception>
         /// <returns><see cref="BlacklistUploadStatus"/></returns>
         Task<BlacklistUploadStatus> PostAsync(string blacklist);
+
+        /// <summary>
+        /// Downloads the blacklist as a tab delimited string, that can be stored as csv file.
+        /// </summary>
+        /// <returns></returns>
```
neilboyd, line 39:
```suggestion
```

Tests/Services/NfieldBlacklistTests.cs
```diff
@@ -42,6 +42,23 @@ public NfieldBlacklistTests()
             _target.InitializeNfieldConnection(mockedNfieldConnection.Object);
         }
 
+        #region GetAsync
+
+        [Fact]
+        public void TestGetAsync_SurveyHasSample_ReturnsSample()
```
neilboyd, line 48: this can also be async/await

gkappen (approved):

neilboyd (approved):

PR #365, opened 2022-11-02
Purple/wait for activity

pilarodriguez:
Wait for the activity to finish and return the download URL (same as we do for the survey data download and many other activities)


Library/Services/INfieldSurveySampleDataService.cs
```diff
@@ -27,5 +27,10 @@ public interface INfieldSurveySampleDataService
         /// Start a new download sample data task, gets the task
         /// </summary>
         Task<BackgroundTask> GetAsync(string surveyId, string fileName);
+
+        /// <summary>
+        /// Start a new download sample data activity, gets the activity
```
evadelnuevo, line 32:
```suggestion
        /// Start a new download sample data activity, gets the activity id
```

evadelnuevo (approved):

PR #367, opened 2022-11-10
add post endpoint in SDK

mvdr:


Library/Services/Implementation/NfieldMediaFilesService.cs
```diff
@@ -110,6 +112,30 @@ public Task AddOrUpdateAsync(string surveyId, string fileName, byte[] content)
                       .FlattenExceptions();
         }
 
+        public Task UploadAndSaveAsync(string surveyId, string fileName, byte[] content)
+        {
+            if (string.IsNullOrEmpty(surveyId))
```
mvdr, line 117: Most services use a private Check method for checking the arguments. This would make the code a bit more readable.
See for instance NfieldSurveySampleDataService => CheckRequiredArgument method,
or in lots of services => CheckSurveyId method.

Library/Services/Implementation/NfieldMediaFilesService.cs
```diff
@@ -110,6 +112,30 @@ public Task AddOrUpdateAsync(string surveyId, string fileName, byte[] content)
...
+                throw new ArgumentNullException("content");
+            }
+
+            var postContent = new ByteArrayContent(content);
+            postContent.Headers.ContentType =
+                new MediaTypeHeaderValue("application/octet-stream");
+
+            return
+               Client.PostAsync(MediaFilesApi(surveyId, fileName), postContent)
+                     .FlattenExceptions();
```
mvdr, line 136: We need to make this proper async/await code. So the method should get the async identifier:
public async Task (string surveyId, string fileName, byte[] content)

and the last line should be:
await Client.PostAsync(MediaFilesApi(surveyId, fileName), postContent).FlattenExceptions().ConfigureAwait(false); 

I am not even sure if FlattenExceptions and ConfigureAwait are necessary.

mvdr (commented):
You forgot to add the tests.
Take a look at NfieldMediaFilesServiceTests.cs file, and do the same thing that was done to test the  AddOrUpdateAsync call.

andreFkantar (approved):

PR #384, opened 2023-01-10
Add theme binary

WimDoze:
[AB#122301](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/122301)

Add possibility to add theme content as a byte array

### DIFF
--- a/Library/Services/Implementation/NfieldThemesService.cs
+++ b/Library/Services/Implementation/NfieldThemesService.cs
@@ -43,20 +43,22 @@ namespace Nfield.Services.Implementation
             var fileName = Path.GetFileName(filePath);
 
             if (!File.Exists(filePath))
-                throw new FileNotFoundException(fileName);
-
-            using (var byteArrayContent = new ByteArrayContent(File.ReadAllBytes(filePath)))
             {
-                byteArrayContent.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
-                var uri = GetUploadThemeUri(templateId, themeName);
-                _ = await Client.PutAsync(uri, byteArrayContent)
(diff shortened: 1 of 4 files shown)


Library/Services/Implementation/NfieldThemesService.cs
```diff
@@ -43,20 +43,22 @@ public async Task UploadThemeAsync(string templateId, string themeName, string f
...
+            await DoUploadThemeAsync(templateId, themeName, themeContent);
+        }
+
+        public async Task UploadThemeAsync(string templateId, string themeName, byte[] themeContent)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(templateId, nameof(templateId));
+            Ensure.ArgumentNotNullOrEmptyString(themeName, nameof(themeName));
+            Ensure.ArgumentNotNull(themeContent, nameof(themeContent));
+
+            await DoUploadThemeAsync(templateId, themeName, themeContent).ConfigureAwait(false);
```
mvdr, line 61: I noticed that all over the SDK we are very inconsistent with using ConfigureAwait. It seems to work with and without it.
Is there a good reason to use it?

mvdr (approved):

neilboyd:
Maybe a `Stream` would be even better than a byte array?

Library/Services/Implementation/NfieldThemesService.cs
WimDoze: I did it here because it was already done for this class.
But I think it is good practice in a library where you do not know where it will be consumed.
When you would for example call this from a windows forms application and would not use configureawait(false) everything would be executed on the UI thread.

WimDoze:
> 

Changed it to a stream

PR #398, opened 2023-03-14
get and set owner as part of the general settings

pilarodriguez:


Library/Models/SurveyGeneralSetting.cs
```diff
@@ -19,6 +19,37 @@ namespace Nfield.Models
...
+
+        /// <summary>
+        /// Indicates whether the survey will be excluded from automatic deletion and stoppage
+        /// </summary>
+        public bool? ExcludeFromAutomaticCleanup { get; set; }
+
+        /// <summary>
+        /// The user id of the owner of the survey 
+        /// </summary>
+        public string Owner { get; set; }
```
hnuguse, line 46: Maybe we should rename this to OwnerId if we are already changing the api

hnuguse (approved):

PR #402, opened 2023-03-27
Removed deprecated endpoints

kefishev:
Removed deprecated calls from services interface

### DIFF
--- a/Examples/ConsoleApplication/Program.cs
+++ b/Examples/ConsoleApplication/Program.cs
@@ -13,15 +13,14 @@
 //    You should have received a copy of the GNU Lesser General Public License
 //    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
 
+using Nfield.Infrastructure;
+using Nfield.Models;
+using Nfield.Services;
 using System;
-using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
-using Nfield.Infrastructure;
-using Nfield.Models;
-using Nfield.Services;
 
 namespace ConsoleApplication
 {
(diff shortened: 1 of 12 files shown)


Examples/ConsoleApplication/Program.cs
```diff
@@ -160,42 +159,6 @@ static void Main()
             surveyFieldworkService.StartFieldworkAsync(newSurvey.SurveyId).Wait();
             surveyFieldworkStatus = surveyFieldworkService.GetStatusAsync(newSurvey.SurveyId).Result; //Should be started
 
-            // Example of a download data request: filtering testdata collected today
-            var surveyDataService = connection.GetService<INfieldSurveyDataService>();
```
pilarodriguez, line 164: as there is a new call for the data download, it would be nice to replace this with the new call instead of deleting it

kefishev: No one is using this example. It ia also not complete, do we want to add all api calls to this?

pilarodriguez: How do you know that no other developers or customers are using this (this is a public API)? We don't have all API calls, but the basic flow. create survey, add interviewer, add sample data, download survey and sample data...
If the example is not used, we should ask others to double check and fully delete the file. But I don't see the reason for deleting just that part, if we can still download survey data with the new endpoint

kefishev: If think so, I will add a download example.

pilarodriguez (commented):
In this repo we also have a postman collections with all the API calls. We need to remove them also from there. It can be done in a separate task

erfosa:
Hi Team,FYI, I am using this API. And ran into difficulties lately using an API endpoint for Download which has not been marked as deprecated/obsolete, but just returned "Access denied". Took me a while to figure this ***@***.***)
-------- Ursprüngliche Nachricht --------Von: pilarodriguez ***@***.***> Datum: 27.03.23  12:47  (GMT+01:00) An: NIPOSoftware/Nfield-SDK ***@***.***> Cc: Subscribed ***@***.***> Betreff: Re: [NIPOSoftware/Nfield-SDK] Removed deprecated endpoints (PR #402) 
@pilarodriguez commented on this pull request.



In Examples/ConsoleApplication/Program.cs:
> @@ -160,42 +159,6 @@ static void Main()
             surveyFieldworkService.StartFieldworkAsync(newSurvey.SurveyId).Wait();
             surveyFieldworkStatus = surveyFieldworkService.GetStatusAsync(newSurvey.SurveyId).Result; //Should be started
 
-            // Example of a download data request: filtering testdata collected today
-            var surveyDataService = connection.GetService<INfieldSurveyDataService>();

How do you know that no other developers or customers are using this (this is a public API)? We don't have all API calls, but the basic flow. create survey, add interviewer, add sample data, download survey and sample data...
If the example is not used, we should ask others to double check and fully delete the file. But I don't see the reason for deleting just that part, if we can still download survey data with the new endpoint

kefishev:
> In this repo we also have a postman collections with all the API calls. We need to remove them also from there. It can be done in a separate task

Created a task

pilarodriguez (approved):

PR #412, opened 2023-04-21
Get and store data retention settings

hnuguse:
- Update SDK
- Update Postman collection

### DIFF
--- a/Library/Models/DataRetentionSettings.cs
+++ b/Library/Models/DataRetentionSettings.cs
@@ -1,0 +1,33 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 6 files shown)


Library/Models/DataRetentionSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The survey data retention period in days
+        /// </summary>
+        public int RetentionPeriod { get; set; }
+
+        /// <summary>
+        /// The possible survey data retention setting values in days
+        /// </summary>
+        public int[] PossibleValues { get; set; }
```
evadelnuevo, line 31: The user should not set these values, no?

evadelnuevo (approved):

Library/Models/DataRetentionSettings.cs
hnuguse: That's right! I changed it to return a fixed list of possible values now.

PR #418, opened 2023-05-10
First blueprint endpoint

joukevandermaas:


Library/Services/Implementation/NfieldSurveysService.cs
```diff
@@ -549,6 +574,11 @@ private static string SamplingPointsQuotaControllerName
             get { return "QuotaTargets"; }
         }
 
+        private static string CreateFromBlueprint
+        {
+            get { return "CreateFromBlueprint"; }
+        }
+
```
MarcoKramer, line 581: Where is this used?

PR #421, opened 2023-05-24
New quota frame endpoint 

AlbertoBelver:
[AB#126608](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/126608)

PR related: 
https://github.com/NIPOSoftwareBV/nfield-source/pull/8350

### DIFF
--- a/Library/Models/SurveyQuotaFrameModels.cs
+++ b/Library/Models/SurveyQuotaFrameModels.cs
@@ -1,0 +1,210 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using System.Collections.Generic;
+
+namespace Nfield.SDK.Models
+{
+
+    /// <summary>
+    /// Represents survey quota frame.
+    /// </summary>
+    public class SurveyQuotaFrameModel
+    {
+        /// <summary>
+        /// Returns the unique identifier of the quota frame. Ignored on the request call
+        /// </summary>
+        public string Id { get; set; }
+
+        /// <summary>
+        /// Returns the unique identifier of the quota version. Ignored on the request call
+        /// </summary>
+        public long? QuotaETag { get; set; }
+
+        /// <summary>
+        /// Global target of the survey quota frame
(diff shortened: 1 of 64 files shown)


Library/Models/SurveyQuotaFrameModels.cs
```diff
@@ -0,0 +1,210 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Nfield.SDK.Models
+{
+
+    /// <summary>
+    /// Represents survey quota frame.
+    /// </summary>
+    public class SurveyQuotaFrameModel
```
joukevandermaas, line 25: I don't think we should duplicate all of these. The explicit intent was to make the quota structure models reusable in the `Nfield.Quota` package (also open source). Any changes should be made there and the models from there should be reused.

joukevandermaas: I don't really understand what you mean. The models in Nfield.Quota package are the new ones, used by the manager, public api, interviewing and SDK. They are not legacy as they are used everywhere. I'd really strongly suggest continuing to use them and not duplicate this code. Looking at the models you added they are 1-for-1 the same.

I know there are some legacy models that use `attributes` etc, those are not the ones I mean.

AlbertoBelver: I'll check it again

AlbertoBelver: There are many differences in the Nfield. Quota models:
- Variables have Name property that really is not used because the name is in the definition.
- Does not have Etag that is available in the Glu API responses
- We cannot use QuotaFrameLevel in the UpdateQuotaTarget because has Children variables and properties that can be updated with the  Glu API endpoint.

We should change a lot of things in Nfield. Quota models to really fit with the things that Glu API allows.
Because this reason I'll keep it as it is. This doesn't mean that we will create a story to review this because at the end the models that we are using internally are the ones defined in:
**Nfield.Manager.Surveys.Interactions.Survey.Quota.Editing**
_All SDK SurveyQuotaFrameModels.cs are similar to .Survey.Quota.Editing_

Tests/Services/NfieldSurveyQuotaFrameServiceTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                                    },
+                VariableDefinitions = new List<SurveyQuotaVariableDefinitionModel>()
+                {
+                        new SurveyQuotaVariableDefinitionModel() { Id = VariableDefinitionId }
+                }
+            };
+        }
+
+        [Fact]
+        public async Task TestQuotaQueryAsync_ServerReturnsQuery()
```
pilarodriguez, line 82: the server returns the quota frame, not the query

Tests/Services/NfieldSurveyQuotaFrameServiceTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        {
+            var ex = await Assert.ThrowsAsync<ArgumentException>(async () => { _ = await _target.CreateOrUpdateQuotaAsync(surveyId, new SurveyQuotaFrameModel()); });
+
+            Assert.Equal("surveyId cannot be null or empty", ex.Message);
+        }
+
+
+
+
+
```
pilarodriguez, line 175: remove the extra lines

pilarodriguez (approved):

PR #438, opened 2023-07-14
Remove CopyableSurveyConfiguration from AddFromBlueprintAsync

MarcoKramer:
---

@NIPOSoftwareBV/team-red

### DIFF
--- a/Library/Services/INfieldSurveysService.cs
+++ b/Library/Services/INfieldSurveysService.cs
@@ -42,7 +42,7 @@ namespace Nfield.Services
         /// <summary>
         /// Adds a new survey based on a blueprint survey.
         /// </summary>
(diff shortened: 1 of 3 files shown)


MartinRijks (approved):
Seems OK but I do wonder if this SDK was ever used, because this would be a breaking change for the people who consumed this method.

But, the earlier the better I suppose.

MarcoKramer:
> Seems OK but I do wonder if this SDK was ever used, because this would be a breaking change for the people who consumed this method.
> 
> But, the earlier the better I suppose.

Yeah, I know. But it's still in early beta, right? And how it worked was not how it was discussed earlier, so tough cookie, I guess. 🤷

MartinRijks:
> Yeah they'll live :D

PR #469, opened 2024-01-19
Single sample record creation with auto generated respondentKey

jlKampos:
[AB#124388](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/124388)

Related PRs:
https://github.com/NIPOSoftwareBV/project-glu-api/pull/5864
https://github.com/NIPOSoftwareBV/nfield-source/pull/9062

### DIFF
--- a/Library/Models/SampleColumnCreate.cs
+++ b/Library/Models/SampleColumnCreate.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Nfield.SDK.Models
+{
+    public class SampleColumnCreate
+    {
+        /// <summary>
+        /// The name of the column
+        /// </summary>
+        public string ColumnName { get; set; }
+
+        /// <summary>
+        /// The value that the column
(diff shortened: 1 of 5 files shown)


Library/Models/SampleColumnCreate.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Nfield.SDK.Models
+{
+    public class SampleColumnCreate
```
ccoceraperez, line 7: Adding the action/verb to the model name is a convention that you follow.? It looks bit odd to me not to have just the name `SampleColumn`. Im checking the target API in case that the model naming is part of the API contract

ccoceraperez: Ok found a SampleColumnUpdate in glu API so I guess that there's a contract somewhere.

ccoceraperez (approved):

PR #473, opened 2024-02-05
Add missing RepositorySurvey endpoints

dstoioskantar:
Related story: [AB#141109](https://dev.azure.com/niposoftware/Nfield/_workitems/edit/141109)

**Description**
Add missing RepositorySurvey endpoints in `NfieldDeliveryRepositorySurveysService`

**Implementation**
- [x] Add missing endpoints

**Tests** 
- [x] Integration tests have been performed

### DIFF
--- a/Library/Services/INfieldDeliveryRepositorySurveysService.cs
+++ b/Library/Services/INfieldDeliveryRepositorySurveysService.cs
@@ -43,7 +43,7 @@ namespace Nfield.Services
         /// </summary>
         /// <param name="repositoryId">The repository id.</param>
         /// <param name="surveyId">The id of the survey to be reinitiated</param>
-        Task PutReinitiateAsync(long repositoryId, string surveyId);
+        Task ReinitiateAsync(long repositoryId, string surveyId);
 
         /// <summary>
         /// Deletes the selected survey from the repository.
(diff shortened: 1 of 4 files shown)


pilarodriguez (commented):
You also need to update the version in the `version.txt` file

Library/Services/INfieldDeliveryRepositorySurveysService.cs
```diff
@@ -30,5 +30,26 @@ public interface INfieldDeliveryRepositorySurveysService
...
+        /// <param name="repositoryId">The repository id.</param>
+        /// <param name="nfieldSurveyIds">The surveys to add</param>
+        Task PostAsync(long repositoryId, string[] nfieldSurveyIds);
+
+        /// <summary>
+        /// Reinitiates a paused Survey in the selected Repository.
+        /// </summary>
+        /// <param name="repositoryId">The repository id.</param>
+        /// <param name="surveyId">The id of the survey to be reinitiated</param>
+        Task PutReinitiateAsync(long repositoryId, string surveyId);
```
gkappen, line 46: Shouldn't this be called `ReinitiateAsync`? The `Put` seems to be an implementation detail on he API that does not have to leak here

dstoioskantar: It was added as it is like that also in the Public Api but it can be omitted

PhilipLipnitskiNipo (approved):

pilarodriguez (approved):

PR #483, opened 2024-03-13
AB#136015: Start interview simulation from Public API

pilarodriguez:
[AB#136015](https://dev.azure.com/niposoftware/Nfield/_sprints/taskboard/Team%20Purple/Nfield/Release%202024/2024%20P1S5?workitem=136015)


Library/Models/InterviewSimulation.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Sample data file name
+        /// </summary>
+        public string SampleDataFileName { get; set;}
+
+        /// <summary>
+        /// Sample data
+        /// </summary>
+        public byte[] SampleDataFile { get; set; }
```
evadelnuevo, line 58: Why the sample data file is received as a byte array and the hints file as a string? Also in the comment of this class it says "To specify uploaded files this model uses the file name and the file content as a text string"

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        public NfieldSurveyInterviewSimulationService(IFileSystem fileSystem)
+        {
+            _fileSystem = fileSystem;
+        }
+
+        /// <summary>
+        /// <see cref="INfieldSurveyInterviewSimulationService.GetHintsAsync(string)"/>
+        /// </summary>
+        public async Task<string> GetHintsAsync(string surveyId)
```
evadelnuevo, line 49: Maybe you can also check here that the surveyId is not null or empty (just for consistency with the other methods)

Tests/Services/NfieldSurveyInterviewSimulationTests.cs
```diff
@@ -0,0 +1,384 @@
...
+            _mockedFileSystem.Setup(fs => fs.File.Exists(SampleDataFilePath)).Returns(false);
+
+            var exception = await Assert.ThrowsAsync<FileNotFoundException>(
+                async () => await _target.StartSimulationAsync(_surveyId, new InterviewSimulationFiles { HintsFilePath = HintsFilePath, SampleDataFilePath = SampleDataFilePath }));
+
+            Assert.Equal(SampleDataFileName, exception.Message);
+        }
+
+        [Fact]
+        public async Task TestStartSimulation_File_HitsAndSampleDataFromFile_PostsExpectedData()
```
evadelnuevo, line 274:
```suggestion
        public async Task TestStartSimulation_File_HintsAndSampleDataFromFile_PostsExpectedData()
```

Library/Models/InterviewSimulation.cs
kefishev: This is because the sample data could use the binary format like xlsx, while the hints are always text.

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
kefishev: checked

evadelnuevo (approved):

PR #485, opened 2024-03-19
Added start simulation functionality

kefishev:
Functions to start interview simulation using different models. One can read data directly form the file path, the other one requires the file name and the file content as a string.
Uinit tests will follow. 

---

@NIPOSoftwareBV/team-blue

### DIFF
--- a/Library/Models/InterviewSimulation.cs
+++ b/Library/Models/InterviewSimulation.cs
@@ -18,7 +18,7 @@ namespace Nfield.Models
     /// <summary>
     /// Parameters to start interview simulation.
     /// To specify uploaded files this model uses the file name and the file content as a text string.
-    /// To upload data form a file path use <see cref="InterviewSimulationFiles"/>
+    /// To upload data directly from files use <see cref="InterviewSimulationFiles"/>
     /// </summary>
     public class InterviewSimulation
     {
--- a/Library/Models/InterviewSimulation.cs
+++ b/Library/Models/InterviewSimulation.cs
@@ -1,0 +1,60 @@
+﻿//    This file is part of Nfield.SDK.
+//
+//    Nfield.SDK is free software: you can redistribute it and/or modify
+//    it under the terms of the GNU Lesser General Public License as published by
+//    the Free Software Foundation, either version 3 of the License, or
+//    (at your option) any later version.
+//
+//    Nfield.SDK is distributed in the hope that it will be useful,
+//    but WITHOUT ANY WARRANTY; without even the implied warranty of
+//    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//    GNU Lesser General Public License for more details.
+//
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace Nfield.Models
+{
+    /// <summary>
+    /// Parameters to start interview simulation.
+    /// To specify uploaded files this model uses the file name and the file content as a text string.
+    /// To upload data form a file path use <see cref="InterviewSimulationFiles"/>
+    /// </summary>
+    public class InterviewSimulation
+    {
+        /// <summary>
+        /// Number of interviews to be simulated
+        /// </summary>
+        public int InterviewsCount { get; set; }
+
+        /// <summary>
+        /// Enable reporting for the simulation
+        /// </summary>
+        public bool EnableReporting { get; set; }
+
+        /// <summary>
+        /// Use sample data from the original survey
+        /// </summary>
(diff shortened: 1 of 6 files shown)


Library/Models/InterviewSimulation.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Use sample data from the original survey
+        /// </summary>
+        public bool UseOriginalSample { get; set; }
+
+        /// <summary>
+        /// Hints file name
+        /// </summary>
+        public string HintsFileName { get; set; }
```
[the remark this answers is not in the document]
kefishev: The file name we need to send it along with the request. The file content should be copied to the Hints field.

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
```diff
@@ -57,6 +134,35 @@ private Uri SurveySimulationHintsEndPoint(string surveyId)
...
+            };
+
+            if (!string.IsNullOrEmpty(simulationRequest.HintsFileName) && !string.IsNullOrEmpty(simulationRequest.Hints))
+            {
+                content.Add(new StringContent(simulationRequest.Hints), nameof(simulationRequest.Hints), simulationRequest.HintsFileName);
+            }
+
+            if (!string.IsNullOrEmpty(simulationRequest.SampleDataFileName) && !string.IsNullOrEmpty(simulationRequest.SampleData))
+            {
+                content.Add(new StringContent(simulationRequest.SampleData), nameof(simulationRequest.SampleData), simulationRequest.SampleDataFileName);
```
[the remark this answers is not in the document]
kefishev: I don't know if it works, did not test it yet.

Library/Services/INfieldSurveyInterviewSimulationService.cs
```diff
@@ -29,5 +30,20 @@ public interface INfieldSurveyInterviewSimulationService
...
+        /// <param name="simulationRequest">Parameters to start the simulations</param>
+        Task StartSimulationAsync(string surveyId, InterviewSimulation simulationRequest);
+
+        /// <summary>
+        /// Starts interview simulation.
+        /// This one allows to upload  hints and sample data directly from files.  
+        /// </summary>
+        /// <param name="surveyId">The survey id</param>
+        /// <param name="simulationRequest">Parameters to start the simulations</param>
+        Task StartSimulationAsync(string surveyId, InterviewSimulationFiles simulationRequest);
```
pilarodriguez, line 47: It should at least return the id of the simulation survey. And probably also the error messages (if any)

kefishev: Is the idea to call from this method the BackgroundActivityController to get the activity result and return it?

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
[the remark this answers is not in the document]
kefishev: This seems to work. I had to change a few model property names.

Library/Services/INfieldSurveyInterviewSimulationService.cs
kefishev: This method can also return the activity id. And the call to BackgroundActivityController could be done separately. But I don't see a service for background activity in the sdk.

kefishev: It appears that the activity status is already available. No need to call the background activity controller, just read the result from with GetActivityResultAsync

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
```diff
@@ -34,6 +55,65 @@ public async Task<Uri> GetHintsAsync(string surveyId)
...
+        /// <see cref="INfieldSurveyInterviewSimulationService.StartSimulationAsync(string, InterviewSimulation)"/>
+        /// </summary>
+        public Task<InterviewSimulationResult> StartSimulationAsync(string surveyId, InterviewSimulation simulationRequest)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(surveyId, nameof(surveyId));
+            Ensure.ArgumentNotNull(simulationRequest, nameof(simulationRequest));
+
+            var content = MultipartDataContent(simulationRequest);
+
+            return Client.PostAsync(StartInterviewSimulationsEndPoint(surveyId), content)
```
pilarodriguez, line 68: we want to move away from this syntax. It is better to keep it as you had it in previous commits (using awaits)

kefishev: Do you mean not use ContinueWith?

pilarodriguez (approved):

PR #493, opened 2024-03-25
Update Nfield Public API.postman_collection.json

jlKampos:


Postman/Nfield Public API.postman_collection.json
```diff
@@ -8846,6 +8846,68 @@
...
+															},
+															{
+																"key": "Authorization",
+																"value": "Basic {{AuthenticationToken}}"
+															}
+														],
+														"body": {
+															"mode": "file",
+															"file": {
+																"src": "/C:/Users/Luciano/Downloads/testing.zip"
```
ccoceraperez, line 8881: This points to your local file path. Will it break if I pull this to my comp and load the postman request or it will simply leave the file value empty for me to search for it?

ccoceraperez (approved):

Postman/Nfield Public API.postman_collection.json
ccoceraperez: Actually I'm goign to try it

jlKampos: > This points to your local file path. Will it break if I pull this to my comp and load the postman request or it will simply leave the file value empty for me to search for it?

wont brake, if you try to upload the file will be missing you need to poin to to annew one

ccoceraperez: Yeah I saw that it does not break, but looks a bit weird that you have an unexistent file loaded in that POSTMAN request body instead of having to select one directly. With this change everybody will have to:

1. Delete the wrong one
2. Select a file

With an empty file field as I suggested:
1. Select a file

PR #498, opened 2024-04-17
[AB#142671] Interview Simulator: List the simulation surveys from Public Api

mvdr:
AB#142671

### DIFF
--- a/Library/Services/INfieldSurveyInterviewSimulationService.cs
+++ b/Library/Services/INfieldSurveyInterviewSimulationService.cs
@@ -14,7 +14,8 @@
 //    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
 
 using Nfield.Models;
-using System;
+using Nfield.SDK.Models;
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Nfield.Services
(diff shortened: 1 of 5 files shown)


Library/Services/INfieldSurveyInterviewSimulationService.cs
```diff
@@ -31,6 +32,19 @@ public interface INfieldSurveyInterviewSimulationService
         /// <returns>Content of the hints file</returns>
         Task<string> GetHintsAsync(string surveyId);
 
+        /// <summary>
```
tdraganidis, line 35: Nice to have the params and the return comments filled with some content.

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
```diff
@@ -139,6 +166,16 @@ private Uri SurveySimulationHintsEndPoint(string surveyId)
             return new Uri(ConnectionClient.NfieldServerUri, $"surveys/{surveyId}/InterviewSimulations/DownloadHints");
         }
 
+        private Uri InterviewSimulationsEndPoint()
+        {
+            return new Uri(ConnectionClient.NfieldServerUri, "surveys/InterviewSimulations");
```
tdraganidis, line 171: Do we have a naming convention for the endpoints? Like camel case or something else? As `surveys` are starting with small letter but InterviewSimulations with a capital one. Not a problem per se, just for consistency in the code.

tdraganidis (approved):
Easy to follow @mvdr ! Just couple of nitpick comments.

Library/Services/Implementation/NfieldSurveyInterviewSimulationService.cs
mvdr: I think this was a mistake. In WebApiConfig in public api all endpoints are defined with a capital. 
I don't think it is case sensitive, but I'll change it to capital.

PR #512, opened 2024-06-27
Purple/create waves

pilarodriguez:


Library/Services/INfieldWavesService.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public interface INfieldWavesService
+    {
+        /// <summary>
+        /// Gets survey queryable object.
+        /// </summary>
+        Task<IQueryable<Survey>> GetParentSurveyWaves(string parentSurveyId);
+
+        /// <summary>
+        /// Adds a new parent survey.
```
mvdr, line 30: It adds a wave, not a parent survey

Library/Services/INfieldWavesService.cs
```diff
@@ -0,0 +1,34 @@
...
+using Nfield.Models;
+
+namespace Nfield.SDK.Services
+{
+    public interface INfieldWavesService
+    {
+        /// <summary>
+        /// Gets survey queryable object.
+        /// </summary>
+        Task<IQueryable<Survey>> GetParentSurveyWaves(string parentSurveyId);
```
mvdr, line 27: also awaitable method, better name it Async as well.

Tests/Services/NfieldWavesServiceTests.cs
```diff
@@ -0,0 +1,87 @@
...
+                .Returns(CreateTask(HttpStatusCode.OK, new StringContent(JsonConvert.SerializeObject(expectedWavesSurveys))));
+
+            var target = new NfieldWavesService();
+            target.InitializeNfieldConnection(mockedNfieldConnection.Object);
+
+            var actualWaves = target.GetParentSurveyWaves(parentSurveyId).Result;
+
+            Assert.Equal(expectedWavesSurveys[0].SurveyId, actualWaves.ToArray()[0].SurveyId);
+            Assert.Equal(expectedWavesSurveys[1].SurveyId, actualWaves.ToArray()[1].SurveyId);
+            Assert.Equal(2, actualWaves.Count());
```
kefishev, line 62: I would not rely on the order in which the waves returned.

Tests/Services/NfieldWavesServiceTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        [Fact]
+        public void TestWaveAsync_ReturnsSurvey()
```
kefishev, line 66: What does it test? Is it better to call it TestWaveAsync_AddWave_ReturnsWaveSurvey or something like this

Tests/Services/NfieldWavesServiceTests.cs
pilarodriguez: It is a mock. So it will always be returned in the order that we specified

Tests/Services/NfieldWavesServiceTests.cs
kefishev: You may also want to clean up usings

kefishev (approved):

Tests/Services/NfieldWavesServiceTests.cs
mvdr: I agree with Kirill on this, don't use the array index.

mvdr: oh right! ok approved

mvdr (approved):

Tests/Services/NfieldWavesServiceTests.cs
pilarodriguez: I don't see why we shouldn't do it. It is a mock. It will return the list in the order that we have specified. We can't do it for real API automated tests, but for unit test should be fine. That is what we are doing in the other tests

PR #517, opened 2024-07-23
Add EnableRespondentsGateway to survey model

hcsauniere:
[AB#151742](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/151742)

- [x] Add Enable Respondents Gateway to Survey

Main PR:
- https://github.com/NIPOSoftwareBV/project-glu-api/pull/6601

### DIFF
--- a/Library/Models/Survey.cs
+++ b/Library/Models/Survey.cs
@@ -65,5 +65,10 @@ namespace Nfield.Models
         /// True if the survey is a blueprint survey, otherwise false
         /// </summary>
         public bool IsBlueprint { get; set; } = false;
+
+        /// <summary>
+        /// True if the survey can be managed by the Respondents Gateway
+        /// </summary>
+        public bool EnableRespondentsGateway { get; set; } = false;
     }
 }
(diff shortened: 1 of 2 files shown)


Library/Models/Survey.cs
```diff
@@ -65,5 +65,10 @@ public Survey(SurveyType surveyType)
         /// True if the survey is a blueprint survey, otherwise false
         /// </summary>
         public bool IsBlueprint { get; set; } = false;
+
+        /// <summary>
+        /// True if the survey can be managed by the Respondents Gateway
+        /// </summary>
+        public bool EnableRespondentsGateway { get; set; } = false;
```
ccoceraperez, line 72: I'm ok with explicitly initialize the property to false, but should not be needed as it's the default value for bool

ccoceraperez (approved):

Library/Models/Survey.cs
hcsauniere: Yeah. I just left it because IsBlueprint was doing the same (to be consistent). I could have removed both actually.

PR #528, opened 2024-08-21
Parent and Waves Isolated models

AlbertoBelver:
[AB#151199](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/151199)

Related PRs:
https://github.com/NIPOSoftwareBV/nfield-public-api/pull/513
https://github.com/NIPOSoftwareBV/nfield-source/pull/9848

### DIFF
--- a/Library/Services/Implementation/NfieldParentSurveyService.cs
+++ b/Library/Services/Implementation/NfieldParentSurveyService.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Nfield.Infrastructure;
 using Nfield.Models;
+using Nfield.Utilities;
 
 namespace Nfield.Services.Implementation
 {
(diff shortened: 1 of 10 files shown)


Library/Services/Implementation/NfieldParentSurveyService.cs
```diff
@@ -38,6 +38,19 @@ public async Task<Survey> AddParentSurveyAsync(Survey survey)
             return JsonConvert.DeserializeObject<Survey>(stringResponse);
         }
 
+        public async Task<Survey> CopyWaveAsync(ParentSurveyWaveCopy survey)
```
pilarodriguez, line 41: Why is this moved to the parent survey service? the parent survey service should be only about the parent surveys (get/create/update/delete), not about waves. We have the AddWave in the waves service, it is weird that when you create a new blank wave it is done from the NfieldParentSurveyWavesService, but if you create it using the artifacts of another wave, then it is done from the NfieldParentSurveyService

AlbertoBelver: Here we have some doubts with the routing, this enpoint is independent of a parent survey an the path is.
`/v1/ParentSurveys/Waves/
image

In this way could imply security issues because maybe you can create wave copies for a Survey in which you don't have permissions.

Currently, we are discussing if we move this endpoint to, before release it. (Or add it to the DA email)

`/v1/ParentSurveys/{ParentSurveyId}/Waves/{ParentSurveyWaveId}/Copy`

{ParentSurveyId} can be use to validate than the wave belongs to the Parent survey and you have permissions, 
{ParentSurveyWaveId} can be use as the SourceWaveId
Model - could contains only the Name and the description of the new wave.

This follows the RESTful standard identifying each resource with which you are performing the actions.
https://github.com/NIPOSoftwareBV/nfield-public-api/wiki/API-Enpoints-Publishing-Checklist#routing

AlbertoBelver: Another thing that is under discussion with Filip and Jorick is that this kind of things should be defined by POs or together with them, before start to work in a new Public API feature.

pilarodriguez (commented):
do we still maintain the postman collection? if so, it should be updated

AlbertoBelver:
Postman collection will be updated in this way.
https://github.com/NIPOSoftwareBV/nfield-public-api/blob/main/Documentation/public-api-migration-guide.md#export-migrated-controllers-list

pilarodriguez (approved):
remember to update the version

PR #530, opened 2024-08-30
Media download filter

AlbertoBelver:
[AB#151176](https://dev.azure.com/niposoftware/15ce0e91-931d-4fbf-9169-8c3dde412b54/_workitems/edit/151176)

The filters introduces the possibility to download only a subset of created media files

- **IncludeCapturedMediaFiles**  This is the current one that already exists to download all media files
Include in the download the All captured Media Files during the interviews (AudioQuestion, VideoQuestion, PhotoQuestion, AudioSilentRecording)

 - New  **IncludeCapturedAudioSilentRecordingFiles**
    Include Audio Silent Recording captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
 
 - New  **IncludeCapturedAudioQuestionFiles**
    Include Audio captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
 
 - New  **IncludeCapturedVideoQuestionFiles**
    Include Video captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
      
 - New  **IncludeCapturedPhotoQuestionFiles**
   Include Photo captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
   
Related PRs:
https://github.com/NIPOSoftwareBV/project-glu-api/pull/6721
https://github.com/NIPOSoftwareBV/nfield-public-api/pull/529
https://github.com/NIPOSoftwareBV/nfield-source/pull/9867

### DIFF
--- a/Library/Models/SurveyDataRequest.cs
+++ b/Library/Models/SurveyDataRequest.cs
@@ -91,25 +91,25 @@ namespace Nfield.Models
         public bool IncludeParaData { get; set; }
 
         /// <summary>
-        /// Include in the download the All captured Media Files during the interviews (AudioQuestion, VideoQuestion, PhotoQuestion, AudioSilentRecording)
+        /// Include in the download all captured Media files during the interviews (AudioQuestion, VideoQuestion, PhotoQuestion, AudioSilentRecording)
         /// </summary>
         public bool IncludeCapturedMediaFiles { get; set; }
         /// <summary>
-        /// Include Audio Silent Recording captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
+        /// Include in the download Audio Silent Recording captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
         /// </summary>
-        public bool IncludeCapturedAudioSilentRecordingFiles { get; set; } = false;
+        public bool IncludeCapturedAudioSilentRecordingFiles { get; set; }
         /// <summary>
-        /// Include Audio captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
+        /// Include in the download Audio captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
         /// </summary>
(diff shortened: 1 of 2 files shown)


Library/Models/SurveyDataRequest.cs
```diff
@@ -91,10 +91,25 @@ public class SurveyDataRequest
...
         /// <summary>
-        /// Include in the download the captured media files during the interviews
-        /// Previously: DownloadCapturedMedia
+        /// Include in the download the All captured Media Files during the interviews (AudioQuestion, VideoQuestion, PhotoQuestion, AudioSilentRecording)
         /// </summary>
         public bool IncludeCapturedMediaFiles { get; set; }
+        /// <summary>
+        /// Include Audio Silent Recording captured Files during the interviews. This parameter will be ignored if IncludeCapturedMediaFiles is true
+        /// </summary>
+        public bool IncludeCapturedAudioSilentRecordingFiles { get; set; } = false;
```
mvdr, line 100: Is there a difference with added setting = false instead of omitting it? As false is the default for a boolean.

Library/Models/SurveyDataRequest.cs
```diff
@@ -91,10 +91,25 @@ public class SurveyDataRequest
         public bool IncludeParaData { get; set; }
 
         /// <summary>
-        /// Include in the download the captured media files during the interviews
-        /// Previously: DownloadCapturedMedia
+        /// Include in the download the All captured Media Files during the interviews (AudioQuestion, VideoQuestion, PhotoQuestion, AudioSilentRecording)
```
mvdr, line 94: "Include all captured Media files ..." or "Include in the download all captured Media files ..."  would read a lot better.

mvdr (approved):

Library/Models/SurveyDataRequest.cs
AlbertoBelver: It was to remark that is false, but no there are no differences.

PR #534, opened 2024-09-16
Manual tests support

mvdr:
---

@NIPOSoftwareBV/team-purple

### DIFF
--- a/Library/Models/ManualTestSurveyResult.cs
+++ b/Library/Models/ManualTestSurveyResult.cs
@@ -13,20 +13,25 @@
 //    You should have received a copy of the GNU Lesser General Public License
 //    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
 
-using Nfield.SDK.Models;
-using System.Linq;
-using System.Threading.Tasks;
+using System.Collections.Generic;
 
-namespace Nfield.SDK.Services
+namespace Nfield.SDK.Models
 {
-    /// <summary>
-    /// Manage manual tests
-    /// </summary>
-    public interface INfieldManualTestsService
+    public class ManualTestSurveyResult
     {
         /// <summary>
-        /// Get All manual tests
+        /// Survey id of the original survey
         /// </summary>
-        Task<IQueryable<SurveyManualTest>> GetAsync();
+        public string OriginalSurveyId { get; set; }
+
+        /// <summary>
+        /// Survey id of the manual test survey
+        /// </summary>
+        public string SimulationSurveyId { get; set; }
+
+        /// <summary>
+        /// Error messages (if any)
+        /// </summary>
+        public IEnumerable<string> ErrorMessages { get; set; } = new List<string>();
     }
 }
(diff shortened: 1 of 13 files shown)


Library/Models/ManualTestSurveyResult.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Survey id of the original survey
+        /// </summary>
+        public string OriginalSurveyId { get; set; }
+
+        /// <summary>
+        /// Survey id of the manual test survey
+        /// </summary>
+        public string SimulationSurveyId { get; set; }
```
pilarodriguez, line 30: This should be TestSurveyId or ManualTestSurveyId

Library/Models/StartCreateManualTestSurvey.cs
```diff
@@ -0,0 +1,46 @@
...
+//    You should have received a copy of the GNU Lesser General Public License
+//    along with Nfield.SDK.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace Nfield.SDK.Models
+{
+    /// <summary>
+    /// Parameters to start creating a manual test survey.
+    /// To specify uploaded files this model uses the file name and the file content.
+    /// For the sample data we expect binary as an input since it could be csv or Excel data. 
+    /// To upload data directly from files use <see cref="InterviewSimulationFiles"/>
```
pilarodriguez, line 22: I think this should be StartCreateManualTestSurveyFile

Library/Services/INfieldSurveyManualTestsService.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// Get manual tests of a specific survey (normally just one)
+        /// </summary>
+        Task<IQueryable<SurveyManualTest>> GetSurveyManualTestsAsync(string surveyId);
+
+        /// <summary>
+        /// Get manual test of a specific survey
+        /// </summary>
+        Task<SurveyManualTest> GetSurveyManualTestAsync(string surveyId, string manualTestId);
```
pilarodriguez, line 36: I don't understand these endpoints. What are they returning? I would expect that we have 1 endpoint without any parameters to return all the manual test surveys in the domain, and another endpoint that for a given survey id, you return the manual test survey

mvdr: The endpoint to return all manual test surveys in the domain is the GetAsync method in NfieldManualTestsService class. This class is for the one with the given surveyId => GetSurveyManualTestsAsync(string surveyId); It returns an array with normally just one test survey. Then we have another endpoint that returns just the one test survey => GetSurveyManualTestAsync(string surveyId, string manualTestId);    You could say that the last one is not needed. But Team Yellow preferred it that way. It is conform the REST standards.

pilarodriguez: But why we need an endpoint for this? If they want to get the manual test survey, they can use the regular GetSurvey endpoint using the manualTestId as surveyId

mvdr: The GetSurvey endpoint does not return the OriginalSurveyId though. But on the other hand, you would have to know the OriginalSurvey to use the new (maybe not needed) endpoint.

PR #542, opened 2024-10-18
SDK EOS Release

jlKampos:
EOS


dquilong (commented):
The last change in the SDK is from Oct 2. This is part of the previous release, so unless there this has been forgotten during the previous sprint, this is not necessary

JoseCarlosLopezGomez (approved):

PR #554, opened 2025-02-28
Update NfieldSurveyInvitationImagesService.cs

jlKampos:
AB#165197

Enables file content type for SurveysInvitationImagesController, missing binary content type, leading to public API errors


jlKampos:
This was discussed in ameeting, we are not going to change V1

PR #556, opened 2025-04-30
Add NfieldSurveyLandingPageService

LuisMacedoKantar:
https://dev.azure.com/niposoftware/Nfield/_workitems/edit/168352


pilarodriguez (commented):
You need to update the version file (https://github.com/NIPOSoftwareBV/Nfield-SDK/blob/master/version.txt)

Library/Services/Implementation/NfieldSurveyLandingPageService.cs
```diff
@@ -0,0 +1,134 @@
...
+        #region Private Methods
+
+        private INfieldHttpClient Client
+        {
+            get { return ConnectionClient.Client; }
+        }
+
+        private Uri LandingPageApi(string surveyId)
+        {
+            return new Uri(ConnectionClient.NfieldServerUri, $"Surveys/{surveyId}/landingPage/");
```
pilarodriguez, line 104: The url is not `return new Uri(ConnectionClient.NfieldServerUri, $"/landingPage/");`? or is that going to change?
image

LuisMacedoKantar: Yes, we’re going to change that, since that URL wasn’t consistent with the rest.
See here: https://niposoftware.slack.com/archives/C076NJ7HNR0/p1746428126648599
But because this change is necessary, this PR might also need to wait a bit.
I’ll let you know when it’s ready for review.
Sorry for the inconvenience, and thank you!

pilarodriguez (approved):

PR #557, opened 2025-05-07
Add unit tests to new landing page service

LuisMacedoKantar:
add unit test; register the service; fix version;

---

@NIPOSoftwareBV/team-red

### DIFF
--- a/Tests/Services/NfieldSurveyLandingPageServiceTests.cs
+++ b/Tests/Services/NfieldSurveyLandingPageServiceTests.cs
@@ -66,7 +66,7 @@ namespace Nfield.Services
         }
 
         [Fact]
-        public void UploadLandingPageAsync_DoesNotThrow()
+        public async Task UploadLandingPageAsync_DoesNotThrow()
         {
             const string surveyId = "SurveyId";
             const string fileName = "MyLandingPage.zip";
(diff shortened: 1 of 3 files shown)


Tests/Services/NfieldSurveyLandingPageServiceTests.cs
```diff
@@ -0,0 +1,107 @@
...
+                            Status = 2 // Succeeded
+                        }))
+                    }))
+                .Verifiable();
+
+            var target = new NfieldSurveyLandingPageService();
+            target.InitializeNfieldConnection(mockedNfieldConnection.Object);
+
+            var stream = new MemoryStream(content);
+            target.UploadLandingPageAsync(surveyId, fileName, stream).Wait();
```
evadelnuevo, line 102: Is there a way to assert that the file was successfully uploaded? Is there a get landing page endpoint or something like that?

gkappen, line 102: Can't we make the test an `async Task` and use `await` here, instead of `.wait()`?

LuisMacedoKantar: This is a unit test, so no real upload happens, the HttpClient is mocked and we simulate the API responses. The test just ensures that UploadLandingPageAsync completes without throwing. I followed the same pattern used in other existing tests, like this [one](https://github.com/NIPOSoftwareBV/Nfield-SDK/blob/92aca25068459c503e7959950b1088172204d78f/Tests/Services/NfieldThemesServiceTests.cs#L148)

LuisMacedoKantar: Done!

evadelnuevo (approved):

